PR #15, opened 2020-03-10
Updates for policies

plorenz:


gendoc.sh
```diff
@@ -1,16 +1,22 @@
 #!/bin/bash
 
+echo $(alias)
```
dovholuknf: these lines are not needed right? (lines 3 and 18)

plorenz: debugging remnants. removed

gendoc.sh
```diff
@@ -1,16 +1,22 @@
...
+
 shopt -s expand_aliases
 
 if [[ "" = "$DOCFX_EXE" ]]; then
     shopt -s expand_aliases
-    source ~/.bash_aliases
+    if [[ -f "~/.bash_aliases" ]]; then
```
dovholuknf, line 7: thx for this - i had this change elsewhere but haven't merged anything lately

dovholuknf (approved):

PR #30, opened 2020-06-08
Update docfx.json

dovholuknf:
updating the repo link for github. pretty sure this will link you to the page you're viewing.

### DIFF
--- a/docfx_project/docfx.json
+++ b/docfx_project/docfx.json
@@ -87,7 +87,7 @@
         "branch": "master",
         "apiSpecFolder": "apidoc"
       }
-      , "_github_link":"https://github.com/openziti"
+      , "_github_link":"https://github.com/openziti/ziti-doc"
       , "_discourse_link":"https://openziti.discourse.group"
       , "_docfxVersion": "2.x"
     }


docfx_project/docfx.json
```diff
@@ -87,7 +87,7 @@
         "branch": "master",
         "apiSpecFolder": "apidoc"
       }
-      , "_github_link":"https://github.com/openziti"
+      , "_github_link":"https://github.com/openziti/ziti-doc"
```
smilindave26, line 90: Is this the link in the top right?  I kinda like it better going to org root than down into the ziti-doc repo.  The 'Improve this doc' looks right...

dovholuknf: It's an option, sure... Before you would get linked directly to the page you were viewing. I'll leave this open for a bit to see if there's any other opinions in either direction.

smilindave26: 'Improve this doc' already links directly to the page you are viewing.

dovholuknf:
thanks for talking me off that ledge :D where's the facepalm emoji?

PR #51, opened 2021-04-27
add some basic controller API help

qrkourier:


dovholuknf:
I like the idea - for sure. I'm torn because this doesn't seem to be an automated process and I think having "no doc" is actually better than doc that goes stale...   If you could rework this so that it this uses the actual spec from the 'latest checkout' that would be better.

I'm just fearful of this going stale if it's not automated...

dovholuknf (approved):

PR #63, opened 2021-12-03
Add forgotten word

voskampm:
'If you want to a second edge router' should probably be 'If you want to create a second edge router'


dovholuknf (approved):
Thank you!

PR #72, opened 2022-03-07
document API specs

qrkourier:


docfx_project/api/rest/index.md
```diff
@@ -1,28 +1,34 @@
...
+The edge-client API services are used by endpoints to interact with the [Ziti Network](~/ziti/overview.md#overview-of-a-ziti-network).
+These services are necessary to authenticate, enroll, return services and fetch other information related to the
 endpoint's participating in the network.
 
-### Download the Client API Spec
+### Download the Edge-Client API Spec
 
-The client api spec can be downloaded from your controller at the following url (replace "${CONTROLLER_URL}" accordingly)
-[https://${CONTROLLER_URL}/edge/client/v1/docs](https://${CONTROLLER_URL}/edge/client/v1/docs).
+The edge-client API spec for your running controller is published as part of the API itself and may be fetched from [https://${CONTROLLER_ADDRESS}/edge/client/v1/docs](https://${CONTROLLER_ADDRESS}/edge/client/v1/docs). The latest edge-client API spec for is also available from [the Git repo](https://github.com/openziti/edge/blob/main/specs/client.yml).
```
rentallect, line 16: Additionally, if it is software that is consuming the spec, it can be fetched from
`https://raw.githubusercontent.com/openziti/edge/main/specs/client.yml`

rentallect (approved):

dovholuknf:
i do not want to see the direct github links in here... if you run an ancient version of ziti - this is probably/possibly out of date and incorrect.

docfx_project/api/rest/index.md
```diff
@@ -1,28 +1,34 @@
...
-[https://${CONTROLLER_URL}/edge/management/v1/docs](https://${CONTROLLER_URL}/edge/management/v1/docs).
+The management API spec too may be downloaded for your running controller at [https://${CONTROLLER_ADDRESS}/edge/management/v1/docs](https://${CONTROLLER_ADDRESS}/edge/management/v1/docs). The latest management API spec is also available from [the Git repo](https://github.com/openziti/edge/blob/main/specs/management.yml).
+
+## The Fabric Management API
+
+The fabric API is a lesser concern for most users, implementers, and integrators because it is itself implemented by the Ziti Edge and so most fabric-level operations are performed on your behalf in the course of using Ziti.
+
+### Download the Fabric Management API Spec
+
+The fabric API spec may be fetched from [the Git repo](https://github.com/openziti/fabric/blob/main/specs/swagger.yml).
```
dovholuknf: i really think it's just too early to put the fabric stuff in this doc for the same reason i don't think it's a good idea to point to the spec from the repo. it can be old/out of date. i suppose it's fine to mention there's a fabric api - but even that feels just premature to me.  i'd rather put in an issue to document all this better from a more wholistic, comprehensive point of view.

qrkourier: For the reasons mentioned in the disclaimer in this diff I care least about Fabric API spec of the three specs and would not shed a tear if we only advertised the latest released spec from the `main` branch for Edge Client and Management API's respective repos.

qrkourier:
> i do not want to see the direct github links in here... if you run an ancient version of ziti - this is probably/possibly out of date and incorrect.

I don't understand your reservation here. It's clear from the text surrounding the links that the best place to fetch **your** spec is from your running controller's APIs. The **latest** spec is just that, and not relevant to a running installation of OpenZiti.

For my purposes, I almost always need the **latest** spec, and this is an attempt to collect those fragments of info in the docs so I don't have to maintain my own bookmarks or build this in to the `ziti` CLI.

docfx_project/api/rest/index.md
qrkourier: I've commented the link to the published Fabric API spec because it is the least important of the three for the purposes of this document.

qrkourier:
@dovholuknf Try the wording on for size here. You'll see that I've carefully introduced the running controller spec first, followed by a helpful link to the published, latest spec. My goal is to provide a reference for the published information, not to guide anyone away from using the built-in spec in their running instance.

qrkourier:
> @dovholuknf Try the wording on for size here. You'll see that I've carefully introduced the running controller spec first, followed by a helpful link to the published, latest spec. My goal is to provide a reference for the published information, not to guide anyone away from using the built-in spec in their running instance.

@dovholuknf I refrained from merging this approved PR because you expressed reservations. I hope that I have addressed those with my later changes and comments.

dovholuknf:
I have tried to see value, but I continue to see no value in pointing people to the latest spec from this online doc.  If people want the inflight version of the spec, let them go out and find it.

What value does providing a possibly incorrect spec provide? I still can't think of any.

qrkourier:
> What value does providing a possibly incorrect spec provide?

@dovholuknf the latest spec is useful for answering questions like "What is the latest spec if I were to choose to upgrade my controller?"

PR #73, opened 2022-03-10
tell people why

philipleonardgriffiths:


docfx_project/ziti/overview.md
```diff
@@ -2,6 +2,15 @@
 uid: zitiOverview
 ---
 
+# Why?
+
```
dovholuknf, line 6: feels like it needs a setup sentence.  Here's how this appears when rendered.  Feels like it needs a setup before just bullet points. Also bullet one feels marketing-ish. We should focus on the developer not the manager/marketer. Maybe just "You don't want to worry about security and networking. OpenZiti has you covered."

image

docfx_project/ziti/overview.md
```diff
@@ -2,6 +2,15 @@
...
+* You want to provide business value to your project as fast as possible but keep having to worry about security and networking.
+* ‘Secure networking’ is an oxymoron. There are no secure networks. Attacks are coming from the network, even from the local network your computer is on right now!
```
dovholuknf, line 8: replace the 'smart' apostrophes with straight quotes: ".

> Attacks are coming from the network
to?
"Attacks can come from any network"

docfx_project/ziti/overview.md
```diff
@@ -2,6 +2,15 @@
...
+* You cannot trust a close sourced solution and you don't want to roll your own. You want the free and easy button.
+
+Great! You have come to the right place.  OpenZiti is the only open-source solution that can place zero trust connectivity directly into your application, making the security of the network entirely irrelevant - internet, local, even host OS. Stop worrying about security and the network and just write your code. Adopt an OpenZiti SDK into your application and you can do just that.
```
dovholuknf, line 11: I'd just eliminate: "Great! You have come to the right place.  "

docfx_project/ziti/overview.md
philipleonardgriffiths: - Maybe go with 'Why did we create Ziti'?
- For the bullets, each should be independent of Ziti, it's the why we created it in the first place. Thus maybe update to 'You want to focus on writing code, not worrying about security and networking' or 'You don't want to worry about security and network, you just want to focus on writing code'
- Finally, I was actually hoping it would go to this page - https://openziti.github.io/ ... I could see why its valuable on both pages.

PR #92, opened 2022-04-29
Fixed: Grammatical error in Service Configuration Overview

BalliAsghar:
A grammatical error has been fixed in Service Configuration Overview


dovholuknf (commented):
I think you want that to read: "can have" - right?

dovholuknf:
Thanks for the fix!

PR #121, opened 2022-06-14
start documenting ziti-edge-tunnel and refreshing the alternative tun…

qrkourier:
…nelers section of the Linux tunnelers article

### DIFF
--- a/docfx_project/ziti/clients/linux.md
+++ b/docfx_project/ziti/clients/linux.md
@@ -38,7 +38,27 @@ RestartSec=3
 WantedBy=multi-user.target
 ```
 
+Where you have created a directory like this.
 
+```bash
+$ tree /opt/openziti
+/opt/openziti
+├── bin
+│   └── ziti-edge-tunnel
+└── etc
+    ├── identities
+    │   └── ACMEIdent.json
+    └── ziti-edge-tunnel.env
+```
+
+And have a unit environment file like this.
+
+```bash
+$ cat /opt/openziti/etc/ziti-edge-tunnel.env
+ZITI_IDENTITY_DIR='/opt/openziti/etc/identities'
+ZITI_DNS_IP_RANGE='100.64.0.1/10'
+ZITI_VERBOSE=3
+```
 
 ### Enroll Before You Run
 
--- a/docfx_project/ziti/clients/linux.md
+++ b/docfx_project/ziti/clients/linux.md
@@ -111,7 +111,7 @@ If any interfaces have a wildcard routing domain configured, `ziti-edge-tunnel` 
 
 ### Run-Host Mode
 
-`ziti-edge-tunnel run-host` provides a narrow subset of the capabilities of the main run mode. This mode only listens for the Services that are authorized to bind to the identities that are loaded.
+`ziti-edge-tunnel run-host` is a mode which is does not allow for service intercepts. The nameserver feature will not be enabled and services which require the intercept functionality will not be available. Only services configured for 'Bind' will be available.
 
 #### Run-Host Mode System Requirements
 
--- a/docfx_project/ziti/clients/linux.md
+++ b/docfx_project/ziti/clients/linux.md
@@ -56,11 +56,6 @@ You will need the token file or its contents to enroll. Enrollment is the act of
 --refresh N
 ```
 
-```bash
-# Specify the tun interface address and the subnet to which Service domain names are resolved (default 100.64.0.1/10). The nameserver address is always the tun interface address +1, default is 100.64.0.2.
---dns-ip-range <ip range>
-```
-
 ### Run Mode
 
 `ziti-edge-tunnel run` provides a transparent proxy and nameserver. The nameserver may be configured to be authoritative (the default) or recursive with a command-line option. The OS is automatically configured to treat the nameserver as primary. You may inspect the resulting configuration with these commands.
@@ -71,6 +66,11 @@ resolvectl domain  # inspect the configuration of search domains
 ```
 
 The configured tun device will ordinarily have the wildcard search domain configured, depending on the pre-existing resolver configuration and the version of systemd.
+
+```bash
+# Specify the tun interface address and the subnet to which Service domain names are resolved (default 100.64.0.1/10). The nameserver address is always the tun interface address +1, default is 100.64.0.2.
+--dns-ip-range <ip range>
+```
 
 #### Run Mode System Requirements
 
--- a/docfx_project/ziti/clients/linux.md
+++ b/docfx_project/ziti/clients/linux.md
@@ -6,7 +6,7 @@
 
 The purpose of the tunneler is to configure host access. This means all users and all processes on the host will share the same level of access. This is accomplished by configuring the OS to have an on-board OpenZiti DNS nameserver and IP routes for authorized OpenZiti Services.
 
-#### Installation and Upgrade
+### Installation and Upgrade
 
 [The latest release](https://github.com/openziti/ziti-tunnel-sdk-c/releases/latest/) of `ziti-edge-tunnel` is distributed as a binary executable from GitHub. The upgrade procedure is identical to the installation procedure.
 
@@ -19,7 +19,7 @@ wget -q "https://github.com/openziti/ziti-tunnel-sdk-c/releases/latest/download/
   && ./ziti-edge-tunnel version
 ```
 
-#### Enroll Before You Run
+### Enroll Before You Run
 
 You will need the token file or its contents to enroll. Enrollment is the act of exchanging the token for an identity that is subsequently installed permanently in the filesystem.
 
@@ -33,7 +33,7 @@ You will need the token file or its contents to enroll. Enrollment is the act of
 ./ziti-edge-tunnel enroll --jwt - --identity ./myTunneler.json < ./myTunneler.jwt
 ```
 
-#### Global Options
+### Global Options
 
 
 ```bash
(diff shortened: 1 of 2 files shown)


docfx_project/ziti/clients/linux.md
```diff
@@ -1,59 +1,147 @@
...
-The linux tunneler is capable of operating in both transparent and proxy modes. In transparent mode the tunneler will
-interact with the operating system to establish intercept points. These intercept points correspond directly to the
-service definitions in place.
+### The Tunneler CLI
 
-`ziti-tunnel` supports several intercept modes, which are specified with a sub-command.
-Typically you will run `ziti-tunnel` with the _run_ sub-command, which determines
-the preferred intercept mode based on the kernel drivers that are available on the
-host.
+`ziti-edge-tunnel` is the general purpose tunneler CLI and can also run as a systemd daemon. For the best overall experience, please use the preferred tunneler `ziti-edge-tunnel` described here.
```
dovholuknf: well... I need to figure out if it's "tunneler" (two l's) and "tunneling" (one l) or if "tunneller" is proper. The dictionary one of my editors uses continuously complains about the "one l" variant being misspelled. Word, for example, considers the "one l" variant wrong.  

I'm pretty sure I have tons of doc written with "tunneller" and "tunneling".... 🤔

docfx_project/ziti/clients/linux.md
```diff
@@ -1,59 +1,147 @@
...
 
-When started on a host that lacks __iptables__, the tproxy intercept mode initializer
-fails and `ziti-tunnel` attempts to use the tun intercept mode:
+### Installation and Upgrade
 
-    $ sudo ziti-tunnel run ziti.json
-    [   0.001]    INFO ziti/tunnel/intercept/tproxy.New: tproxy listening on 127.0.0.1:37313
-    [   0.001]    INFO ziti/tunnel/cmd/ziti-tunnel/subcmd.run: tproxy initialization failed: failed to initialize iptables handle: exec: "iptables": executable file not found in $PATH
-    [   0.009]    INFO ziti/tunnel/cmd/ziti-tunnel/subcmd.run: using tun interceptor
+[The latest release](https://github.com/openziti/ziti-tunnel-sdk-c/releases/latest/) of `ziti-edge-tunnel` is distributed as a binary executable from GitHub. The upgrade procedure is identical to the installation procedure.
```
dovholuknf, line 11: Will you file an issue to track when this is part of package managers to come back here and update this with "easier" instructions? this script indicates bash but is it `sh` compatible? Would/should we add a helper function to the ziti-cli-scripts.sh to add getting the linux binary? (this function basically it seems?)

docfx_project/ziti/clients/linux.md
```diff
@@ -1,59 +1,147 @@
...
+```bash
+# shell script illustrating the steps to install or upgrade ziti-edge-tunnel
+wget -q "https://github.com/openziti/ziti-tunnel-sdk-c/releases/latest/download/ziti-edge-tunnel-Linux_$(uname -p).zip" \
+  && unzip ./ziti-edge-tunnel-Linux_$(uname -p).zip \
+  && rm ./ziti-edge-tunnel-Linux_$(uname -p).zip \
+  && chmod -c +x ./ziti-edge-tunnel \
+  && ./ziti-edge-tunnel version
+```
+
+Optionally, you may install `ziti-edge-tunnel.service` as a systemd unit. For example,
```
dovholuknf: I really try not to document HOW to use systems in this way. It's helpful, but it opens us up to some questions from the doc. we do make systems files for the quickstart in the ziti-cli-scripts.sh, maybe this is a function to just add in there and then turn this section into a "source this file and run createSystemdUnitFileForTunneller" (or whatever?)

docfx_project/ziti/clients/linux.md
qrkourier: My guess is that "tunneller" is UK, "tunneler" is US. We can assume our style guide will say "Use US spellings", but we can also have exceptions like "Zitit tunnellers are spelled with two Ls". I could also be guessing wrong about which is the US spelling.

docfx_project/ziti/clients/linux.md
```diff
@@ -1,59 +1,147 @@
...
+Restart=always
+RestartSec=3
+
+[Install]
+WantedBy=multi-user.target
+```
+
+
+
+### Enroll Before You Run
```
dovholuknf, line 22: This should all be a pointer to the enroll doc and this sort of example should be on those pages (if not already)

docfx_project/ziti/clients/linux.md
qrkourier: https://github.com/openziti/ziti-doc/issues/122

qrkourier: This script is `sh` compatible and doesn't use any modern BASHisms. My preference is to see the few lines of clear, pasteable shell code over a comparatively opaque shell function.

docfx_project/ziti/clients/linux.md
```diff
@@ -1,59 +1,147 @@
...
+# enroll from a token file
+./ziti-edge-tunnel enroll --jwt ./myTunneler.jwt --identity ./myTunneler.json
+```
+
+```bash
+# enroll from stdin
+./ziti-edge-tunnel enroll --jwt - --identity ./myTunneler.json < ./myTunneler.jwt
+```
+
+### Global Options
```
dovholuknf, line 28: I kinda thinking just dumping the output the help command here and let it match and then talk about the options below might be a good idea. keeping this up to date though - is challenging. for that reason, I tend to __not__ document things like this that are already mostly documented in the help command...  i could be convinced that it's needed, but it's just hard to support

docfx_project/ziti/clients/linux.md
qrkourier: For myself, I do want convenient helper functions like that in my quickstart, but I oppose the idea of vending functions to users in this context. Wrapper shell should be used judiciously and code-coupled to the thing it's wrapping. In other words, I feel like we should ship helper shell scripts and functions with `ziti-edge-tunnel` if there's no other way to accomplish the goal. The Linux packages for example are the right solution to this problem.

docfx_project/ziti/clients/linux.md
qrkourier: I agree about not documenting redundantly. In this case the global options are not presented as such in the built-in usage hints, and so I think it's expedient to clearly document them at least one place.

docfx_project/ziti/clients/linux.md
```diff
@@ -1,59 +1,147 @@
...
+# Set service polling interval in seconds (default 10).
+--refresh N
+```
+
+### Run Mode
+
+`ziti-edge-tunnel run` provides a transparent proxy and nameserver. The nameserver may be configured to be authoritative (the default) or recursive with a command-line option. The OS is automatically configured to treat the nameserver as primary. You may inspect the resulting configuration with these commands.
+
+```bash
+resolvectl dns     # inspect the association of tun device and nameserver
```
dovholuknf, line 56: this is the first time resolvectl is referenced right? shouldn't it be somewhere in here that it's required?

qrkourier: It's not required any longer by `ziti-edge-tunnel`, but useful for inspecting the configuration. There is yet a fallback behavior in ZET to shell out to that CLI, but only if it was built with a non-default option to disable the default behavior.

docfx_project/ziti/clients/linux.md
```diff
@@ -1,59 +1,147 @@
...
+resolvectl domain  # inspect the configuration of search domains
+```
+
+The configured tun device will ordinarily have the wildcard search domain configured, depending on the pre-existing resolver configuration and the version of systemd.
```
dovholuknf: "ordinarily" reads awkwardly. How about:

> If any interfaces have a wildcard routing domain configured, `ziti-edge-tunnel` will also configure its tun with a wildcard routing domain.  If no other interface has a wildcard routing domain configured, neither with the `ziti-edge-tunnel` tun.

docfx_project/ziti/clients/linux.md
```diff
@@ -1,59 +1,147 @@
...
+```bash
+# Specify the tun interface address and the subnet to which Service domain names are resolved (default 100.64.0.1/10). The nameserver address is always the tun interface address +1, default is 100.64.0.2.
+--dns-ip-range <ip range>
+```
 
-#### tproxy
+#### Run Mode System Requirements
 
-tproxy is the preferred intercept mode when running on a Linux kernel that has the __ip_tables__ kernel module installed.
+`ziti-edge-tunnel run` requires elevated privileges for managing the `/dev/net/tun` device, IP routes, and works best when `libsystemd` is available so that it can call the `systemd` API `/var/run/dbus/system_bus_socket` to configure nameservers and search domains for `systemd-resolved`.
```
dovholuknf: > and works best when `libsystemd` is available so that it can call the `systemd` API `/var/run/dbus/system_bus_socket`

please remove this. it's in the weeds too far. It's better to state the order of precedence which the tunneler operates. It tries to use the systemd API if it's there, then tries to use `resolvectl`, then it falls back to mucking with the resolve.conf file. And when you do that it's not really a 'system requirement' and probably belongs more in the prior section. 

the only 'requirement' I can see is that you really need elevated privs. we should callout CAP_NET_ADMIN or root - assuming CAP_NET_ADMIN is still enough?

docfx_project/ziti/clients/linux.md
```diff
@@ -1,59 +1,147 @@
...
 
-    $ lsmod | grep ip_tables
-    ip_tables              32768  5 iptable_filter,iptable_security,iptable_raw,iptable_nat,iptable_mangle
+### Run-Host Mode
 
-`ziti-tunnel` manipulates routing tables and firewall rules when using the tproxy
-intercept mode. The NET_ADMIN Linux capability is required for these actions. The
-usage example here runs ziti-tunnel with sudo as a simple way to obtain
-NET_ADMIN:
+`ziti-edge-tunnel run-host` provides a narrow subset of the capabilities of the main run mode. This mode only listens for the Services that are authorized to bind to the identities that are loaded.
```
dovholuknf: > provides a narrow subset of the capabilities of the main run mode

is it 'narrow'? To me, it's "half" of the features? What about something like:

> `ziti-edge-tunnel run-host` is a mode which is does not allow for service intercepts. The nameserver feature will not be enabled and services which require the intercept functionality will not be available. Only services configured for 'Bind' will be processed.

docfx_project/ziti/clients/linux.md
dovholuknf: "Run Mode System Requirements" is also redundant - let's just use "System Requirements" since it's nested under the previous section

docfx_project/ziti/clients/linux.md
```diff
@@ -1,59 +1,147 @@
...
-    [   0.010]    INFO ziti/tunnel/dns.NewDnsServer: starting dns server...
-    [   2.018]    INFO ziti/tunnel/dns.NewDnsServer: dns server running at 127.0.0.1:53
-    [   2.018]    INFO ziti/tunnel/dns.(*resolver).AddHostname: adding ziti-tunnel.resolver.test = 19.65.28.94 to resolver
-    [   2.033]    INFO ziti/tunnel/dns.(*resolver).RemoveHostname: removing ziti-tunnel.resolver.test from resolver
-    [   2.096]    INFO ziti/tunnel/cmd/ziti-tunnel/subcmd.updateServices: starting tunnel for newly available service wttr.in
-    [   2.290]    INFO ziti/tunnel/dns.(*resolver).AddHostname: adding wttr.in = 5.9.243.187 to resolver
-    [   2.300]    INFO ziti/tunnel/cmd/ziti-tunnel/subcmd.updateServices: service wttr.in not hostable
-    [   2.300]    INFO ziti/tunnel/cmd/ziti-tunnel/subcmd.updateServices: starting tunnel for newly available service ssh-local
-    [   2.570]    INFO ziti/tunnel/dns.(*resolver).AddHostname: adding local.io = 169.254.1.1 to resolver
+#### Run-Host Mode System Requirements
```
dovholuknf: > Run-Host Mode

imo - redundant as per above

dovholuknf (approved):

docfx_project/ziti/clients/linux.md
qrkourier: Markdown headings should be unique, and so I've used the more specific heading to avoid a collision with the one about Run-Host Mode

docfx_project/ziti/clients/linux.md
qrkourier: Spellcheck agrees that singular is "tunneller."

PR #127, opened 2022-06-24
Fixing zac service code snippet

gberl002:
Fixing zac service code snippet so the formatting is correct when copying and pasting. Also fixed some notes not rendering as expected.

### DIFF
--- a/docfx_project/ziti/quickstarts/zac/installation.md
+++ b/docfx_project/ziti/quickstarts/zac/installation.md
@@ -123,43 +123,43 @@ this time.  It's not difficult to reuse the PKI but you'll need to do the follow
 4. Now add the ZAC configuration lines to the compose file of your choice:
    ```bash
      ziti-console:
-    image: openziti/zac
-    ports:
-      - "1408:1408"
-      - "8443:8443"
-    environment:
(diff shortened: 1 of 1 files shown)


docfx_project/ziti/quickstarts/zac/installation.md
```diff
@@ -123,43 +123,43 @@ this time.  It's not difficult to reuse the PKI but you'll need to do the follow
...
+         - zitired
    ```
 1. After adding the ZAC configuration as shown, `docker-compose` will now start and expose the ZAC ports on 1408/8443.
-   > [!Note]
-   > Do note that if you are exposing ports as shown above, you will need to ensure that `ziti-edge-controller` is
-   > addressable by your machine in order to use docker in this way. This guide does not go into how to do this in depth.
-   > One easy, and common mechanism to do this would be to edit the 'hosts' file of your operating system. A quick
-   > internet search should show you how to accomplish this.
-   > 
+> [!Note]
```
gberl002, line 146: I'm not sure if this was intended but this note is an exact copy of the note just before this section. They have different commit timestamps but I'm assuming that, based on the fact that they are identical and not just closely worded, that this was intended so I'm leaving it in for now.

qrkourier (approved):

PR #159, opened 2022-09-21
crosslink to the Docker tunneler readme

qrkourier:


docusaurus/docs/clients/tunnelers/01-linux.md
```diff
@@ -165,6 +165,10 @@ If the DNS record exists it returns the answer and sets query status to `NO_ERRO
 
 `ziti-edge-tunnel run-host` does not require elevated privileges or the above device or socket, only network egress to the servers for which it is hosting Services.
 
+### [Run with Docker](https://github.com/openziti/ziti-tunnel-sdk-c/tree/main/docker#readme)
+
+The `run-host` mode is especially useful as a sidecar for publishing isolated servers in a Docker bridge network or the Docker host's network.
```
dovholuknf: is "especially" necessary? It seems like it's superfluous to me.  also this whole section is about run-host not "run with docker", it doesn't fit for me. dunno, i don't understand the motivation for including it here. maybe it just needs a whole section of its own and further detail to get whatever point you're thinking of across?

dovholuknf (approved):
it'd be cool to tell people 'why' but -- lgtm :)

PR #195, opened 2022-10-25
Update ziti router

dariuszSki:


qrkourier:
@dariuszSki Thanks for this! I only took a quick look at this so far and noticed:
* Not certain this is relevant in this case, but when possible do use `git mv` to rename a file so that the changes will be evident in the diff
* In this case destinations like /docs/manage/router-overview need to be preserved because there was already an article published to that path, so you could avoid renaming the Markdown file or you could update your branch from main to see the new HTTP redirects that were added in `docusaurus.config.js` and add one there for the relocated file
* there's an unresolved conflict with 03-components.md

dariuszSki:
Sure, I did not expect this version to be merged as is. Just wanted feedback using this PR,  so please review it if you can. I am open to m make this better.  I am already in the process of splitting the content into sub MD files. There is too much information crammed into one MD file.

dariuszSki:
I see what happened here, the components folder structure changed since my last sync and now it is out of sync.  Probably better coordination is required when major changes are done to the structure of the folders and files.

dariuszSki:
@dovholuknf if it is not too much to ask. Just wondering if you are planning to share the content of ```_components.md``` in at least 2 different places?

dovholuknf:
It's referenced from 01-overview.md and 03-components.md now.  I think we should have specific pages for each of those components. I see you're expanding on the components, but i think we'd like to keep the _components.md small for now so that the overview page doesn't get too much busier. If it makes sense, I could see moving the 'overview' type stuff to the overview (the existing content) then expanding on them like you are in the 03-components.md.  That or maybe add the extra details you are in a whole new router-focused md?

dariuszSki:
ok, I see. Imho, we should not repeat  the same exact content in more place than one.  I am also in the camp where the overview would only talk about openziti router at the high level and possibly not to mention fabric and edge at this stage. Then we can provide a link to router concept where we could explain in more details the fabric , edge , tunneler options. Then in manage section explaining the deployment details of these options. Just a suggestion, and I think you are thinking along those lines if I am not mistaking.

dariuszSki:
see new PR https://github.com/openziti/ziti-doc/pull/229, closing this one

Issue #211, opened 2022-10-28
some redirects from old links are not working


qrkourier:
This reflects local testing with my dev server kpop4:3000 which is currently on https://github.com/openziti/ziti-doc/pull/212

```bash
❯ ./check-old-urls.sh kpop4:3000                                               
Failed:
        /ziti-android-app/README.html
        /articles/bootstrapping-trust/part-01.encryption-everywhere.html
        /articles/bootstrapping-trust/part-05.bootstrapping-trust.html
        /articles/bootstrapping-trust/part-02.a-primer-on-public-key-cryptography.html
        /articles/bootstrapping-trust/part-04.certificate-authorities-and-chains-of-trust.html
        /articles/bootstrapping-trust/part-03.certificates.html
```

Also, these destinations no longer exist:

```bash
# /api/csharp/NetFoundry.html
# /api/csharp/NetFoundry.ZitiConnection.html
# /api/csharp/NetFoundry.ZitiIdentity.html
# /api/csharp/NetFoundry.Ziti.html
# /api/csharp/NetFoundry.ZitiConnection.OnConnected.html
# /api/csharp/NetFoundry.ZitiConnection.OnDataReceived.html
# /api/csharp/NetFoundry.ZitiConnection.OnDataWritten.html
# /api/csharp/NetFoundry.ZitiException.html
# /api/csharp/NetFoundry.ZitiStream.html
# /api/csharp/NetFoundry.ZitiStatus.html
```

qrkourier:
For some reason Vercel behaves differently. 

```bash
❯ ./check-old-urls.sh kpop4:3000                                 
Failed:
        /articles/bootstrapping-trust/part-01.encryption-everywhere.html
        /articles/bootstrapping-trust/part-02.a-primer-on-public-key-cryptography.html
        /articles/bootstrapping-trust/part-03.certificates.html
        /articles/bootstrapping-trust/part-04.certificate-authorities-and-chains-of-trust.html
        /articles/bootstrapping-trust/part-05.bootstrapping-trust.html

❯ ./check-old-urls.sh https://ziti-er4yfltbh-openziti.vercel.app
Failed:
        /api/clang/index.html
        /api/swift/index.html
        /articles/bootstrapping-trust/part-01.encryption-everywhere.html
        /articles/bootstrapping-trust/part-02.a-primer-on-public-key-cryptography.html
        /articles/bootstrapping-trust/part-03.certificates.html
[...]

PR #212, opened 2022-10-28
port blog articles to Docusaurus and decommission DocFX

### LINKED ISSUE #194: rip out docfx
as the final step, we need to replace the docfx site entirely with the docusaurus site

qrkourier:
Some redirects not working in https://github.com/openziti/ziti-doc/issues/211


dovholuknf (approved):
so long docfx

qrkourier:
I'm seeing an inconsistency between the way Docusaurus's redirects behave with `yarn serve` (minified prod serve from static `./build`) and Vercel. I'll have to investigate before merging because I'm worried it'll break most of the redirects that are currently working.

PR #215, opened 2022-10-31
Buildfix publish script

qrkourier:


qrkourier:
This change is necessary because publish.sh currently encounters permission denied when it tries to mv the `ziti-ci` executable to root's `/usr/bin`. This succeeded in the DocFX containerized build because the script was running as root inside the container, but now it's running as the Actions user on the hosted runner VM because there's no longer any need for the build container that was decommissioned along with DocFX.

dovholuknf (approved):

qrkourier:
This branch fixed some but not all of the Docusaurus publish script problems. Remaining issues described in https://github.com/openziti/ziti-doc/issues/220

PR #223, opened 2022-11-01
use absolute targets with trailing slashes for the meta refresh redir…

### LINKED ISSUE #222: Some API docs nav links broken
The sidebar nav links in API docs section for csharp and swift get 404 when deployed to GH Pages. The sidebar links are working with `yarn serve` so there must be some difference with GH Pages.

We can't see this problem in Vercel because when we deploy to Vercel we don't build the companion microsites in _remotes that provide this content. We would need to find a way to install Doxygen, an executable used to build the C SDK docs, in the Vercel build env and change the Vercel build command to stop sending the `-l` option to `./gendoc.sh`.

qrkourier:
…ects that populate API Docs sidebar nav links

### DIFF
--- a/docusaurus/docs-api/ziti-c-sdk.mdx
+++ b/docusaurus/docs-api/ziti-c-sdk.mdx
@@ -4,5 +4,5 @@ hide_title: true
 ------
 
 <head>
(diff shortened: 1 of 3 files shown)


qrkourier:
Closing because the problem was resolved without this change.

PR #226, opened 2022-11-04
tweak tagline

qrkourier:


dovholuknf:
https://becomeawritertoday.com/is-with-capitalized-in-a-title/

this was all intentional anyway

qrkourier:
Agreed on capitalizing "With". Now the only change is to strip the leading and trailing `--` characters from the tagline, which I think look ugly.

PR #227, opened 2022-11-04
add a page to document how search works

qrkourier:


docusaurus/src/pages/about-search.md
```diff
@@ -0,0 +1,10 @@
+
+# How Search Works
```
dovholuknf: this seems like meta-doc about the site itself. Why add this to the site itself and not just as a markdown page of the repo itself? I'm not sure this needs to be published?

qrkourier: Agreed. I moved the Algolia search documentation to the main README.

dovholuknf (approved):

PR #229, opened 2022-11-04
new-router-update

### LINKED ISSUE #200: publish Docusaurus site in GitHub pages
Replace the DocFx site that's published in openziti.github.io

dariuszSki:


docusaurus/docs/manage/02-router/01-overview.md
```diff
@@ -0,0 +1,63 @@
...
+</Tabs>
+
+## Sizing Guidelines
+More details can be found [here](03-sizing.md)
+## Process of Deployment
+In order to eliminate sharing private key(s) over internet (i.e. existing key(s) located on a target host), it is best practice to run the cli router creation commands  on the target host. 
+
+:::info Note
+Ensure you are [logged in](../04-cli/02-logging-in.md) 
+for creating/udating routers through Cli
```
qrkourier: updating

docusaurus/docs/manage/02-router/01-overview.md
```diff
@@ -0,0 +1,63 @@
...
+:::
+
+1. ### Download ziti and ziti-router binaries
+    [openziti release notes and artifacts](https://github.com/openziti/ziti/releases)
```
qrkourier: Shall we link to the latest release to facilitate the immediate download of artifacts?
https://github.com/openziti/ziti/releases/latest

docusaurus/docs/manage/02-router/01-overview.md
```diff
@@ -0,0 +1,63 @@
...
+More details can be found [here](03-sizing.md)
+## Process of Deployment
+In order to eliminate sharing private key(s) over internet (i.e. existing key(s) located on a target host), it is best practice to run the cli router creation commands  on the target host. 
+
+:::info Note
+Ensure you are [logged in](../04-cli/02-logging-in.md) 
+for creating/udating routers through Cli
+:::
+
+1. ### Download ziti and ziti-router binaries
```
qrkourier: I think headings alone, without the ordered list, are sufficient here to express the procedure. The formatting is awkward combining the list with heading.

docusaurus/docs/manage/02-router/02-configuration.md
```diff
@@ -0,0 +1,300 @@
...
+
+forwarder:
+    latencyProbeInterval: 10
+    xgressDialQueueLength: 1000
+    xgressDialWorkerCount: 128
+    linkDialQueueLength: 1000
+    linkDialWorkerCount: 32
+```
+</TabItem>
+<TabItem value="Public-Fabric" label="Public Router" attributes={{className: styles.orange}}>
```
qrkourier: Please comment this placeholder so it doesn't show up in the rendering.

docusaurus/docs/manage/02-router/03-sizing.md
```diff
@@ -0,0 +1,229 @@
...
+id: router-sizing
+title: Sizing
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import styles from './styles.module.css';
+
+<Tabs>
+<TabItem value="vcpe" label="vCPE (generic type)" attributes={{className: styles.custom}}>
```
qrkourier: I feel like "vCPE" is a meaningless term in this context, especially if it has to be explained as "generic". How about simply "Generic"?

docusaurus/docs/manage/02-router/04-update/01-cli.md
```diff
@@ -0,0 +1,99 @@
...
+<TabItem value="Public-Edge" label="Public Router with Edge" attributes={{className: styles.orange}}>
+
+```bash
+ziti edge create edge-router $ROUTER_NAME \
+                            --jwt-output-file $ROUTER_NAME.jwt
+```
+
+</TabItem>
+<TabItem value="Public-Fabric" label="Public Router" attributes={{className: styles.orange}}>
+To be added
```
qrkourier: It seems best to comment this until we have something useful to publish.

docusaurus/docs/manage/02-router/04-update/01-cli.md
```diff
@@ -0,0 +1,99 @@
...
+ziti edge create edge-router $ROUTER_NAME \
+                            --jwt-output-file $ROUTER_NAME.jwt
+```
+
+</TabItem>
+<TabItem value="Gateway" label="Private Router with Edge and Tunneler" attributes={{className: styles.green}}>
+
+:::info Notes
+`--no-travesal` flag is not required, but keep in mind that private routers are stub routers and setting it to true disables transitive routing through it.
+In other words, only connections destined for this router will be routed to it by the samrt routing algorithm. `--tunneler-enabled or just -t` flag indicates the tunnel mode.
```
qrkourier: smart routing

docusaurus/docs/manage/04-cli/_login-snippet.md
```diff
@@ -0,0 +1,10 @@
+Load the current user/password into an environment variables
+
+ctrl_user=$(jq -r .username ~/.config/ziti/ziti-controller/credentials.json)
```
qrkourier: Enclose BASH snippet in syntax fences like

    ```bash
    bashisms
    ```

qrkourier: I'm not sure where this file came from `~/.config/ziti/ziti-controller/credentials.json`. If it's not a widely-available, standard file location then this has little value for the reader. Does this need to be updated to use the new `ziti` CLI login profiles?

docusaurus/docusaurus.config.js
```diff
@@ -293,10 +293,6 @@ const config = {
             to: '/docs/core-concepts/metrics/inspect',
             from: ['/ziti/metrics/inspect.html'],
           },
-          {
```
qrkourier, line 296: I'm guessing you removed these because the `to:` destinations were invalid after re-organizing things. If so, then let's correct the value of `to:` instead of deleting the redirect altogether. That'll preserve the old `from:` permalink and keep it pointing at the new, best destination.

qrkourier (changes requested):

docusaurus/docs/manage/02-router/01-overview.md
dariuszSki: ok, will update

docusaurus/docs/manage/02-router/01-overview.md
dariuszSki: ok, will correct

docusaurus/docs/manage/02-router/01-overview.md
dariuszSki: yes probably, thinking was to use heading to create sidebar shortcuts

docusaurus/docs/manage/02-router/02-configuration.md
dariuszSki: yes, I created a placeholder for this not to forget. Before this PR merge was completed, I was going to fill this in once ```ziti create config router edge``` changes were merged into the new ziti release.

docusaurus/docs/manage/02-router/03-sizing.md
dariuszSki: sure, sounds like a good idea

docusaurus/docs/manage/02-router/04-update/01-cli.md
dariuszSki: same as one before

docusaurus/docs/manage/02-router/04-update/01-cli.md
dariuszSki: thanks

docusaurus/docs/manage/04-cli/_login-snippet.md
dariuszSki: sure, it is  probably a good time to update it.

docusaurus/docusaurus.config.js
dariuszSki: Sure, I think this file is not up-to-date either. Not sure how this file is updated, manually?

docusaurus/docs/core-concepts/security/authorization/policies/_create-edge-router-policy-cli.md
```diff
@@ -1,4 +1,4 @@
-To create an Edge Router Policy using the CLI issue the following commands. (ensure you are [logged in](/docs/manage/cli/logging-in))
+To create an Edge Router Policy using the CLI issue the following commands. (ensure you are [logged in](/docs/manage/04-cli/logging-in.md))
```
dovholuknf: with docusaurus, I've found that it's a bitt better in the long run to not use the .md in the link. if the .md is renamed to .mdx it won't break all the links to it.

docusaurus/docs/manage/02-router/01-overview.md
```diff
@@ -0,0 +1,63 @@
+---
+id: router-deploying-overview
+title: Deployment Overview
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import styles from './styles.module.css';
+
+In this article, we are describing the process of router deployment and what attributes can be updated, removed, added after the deployment is completed.
```
dovholuknf: In technical docs, it's better to maintain a more formal tone and avoid informal words like "you", "we", "I" etc. Those sorts of words are better for a blog or tutorial type of doc. This sentence should be reworded:

**From:**
In this article, we are describing the process of router deployment and what attributes can be updated, removed, added after the deployment is completed.

**To:**
This article describes the process of deploying a router. It covers what attributes can be updated, removed and added after the deployment is completed.

docusaurus/docusaurus.config.js
qrkourier: Yep. This is the Docusaurus configuration file. You can rebase on `issue-106-router-docs-overhaul` if you add the upstream repo as a Git remote and fetch it. That'll make sure you have the latest file.

docusaurus/docs/manage/02-router/01-overview.md
```diff
@@ -0,0 +1,63 @@
...
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import styles from './styles.module.css';
+
+In this article, we are describing the process of router deployment and what attributes can be updated, removed, added after the deployment is completed.  
+
+## Components of Deployment
+
+:::info Notes
+Currently, the DNS name or IP address can not be changed after the deployment, because it requires to update the `Sans` section of the router server certificate. To do that one would need to re-enroll the router, which can not be done without creating a new router.
```
dovholuknf: Personally, I would favor leaving out extra details that aren't explicitly necessary like the SANS stuff. Ideally, we'd point people to a link where that whole thing is described but we don't have that yet, so I'd just leave that detail out

**From:**
Currently, the DNS name or IP address can not be changed after the deployment.

**To:**
Changing a router's advertised DNS entry or IP address (if not DNS-based) is not supported. To change these values after enrollment, the router must be deleted and re-enrolled

It's basically the same but doesn't get into the SANS bits which I expect would confuse all but the hardest core of readers

docusaurus/docs/manage/02-router/01-overview.md
```diff
@@ -0,0 +1,63 @@
...
+:::
+
+<Tabs>
+<TabItem value="Identity" label="Identity" attributes={{className: styles.green}}>
```
dovholuknf: I'd really rather see these not presented as tabs so that a reader can just scroll up and down and so that the right nav bar is useful to locate the sections if needed

docusaurus/docs/manage/02-router/01-overview.md
```diff
@@ -0,0 +1,63 @@
...
+The pki-related fields in the `identity` section of the router configuration files are important to understand. These
+files are generated during the process of [enrolling](#enrollment) the router. These files do not need to exist before and will be created during the enrollment process. 
+This means the process running the enrollment will need the correct privileges granted in order to write - or overwrite those files in case of re-enrollment.  
+If the key specified in the identity section already exists - it will not be overwritten. Instead, it will be used during the enrollment process. In other words, you can create your own key or use existing key.
+
+</TabItem>
+<TabItem value="Config" label="Configuration File" attributes={{className: styles.orange}}>
+
+The router is configured using a [yaml](https://yaml.org/) file. The configuration file can be obtained by running the `ziti` CLI binary on the destiantion host, 
+but some configration options will need to be passeed to match the deployment details, i.e hostname, ip, private or public routers, etc. See `Process of deployment` for more.
```
dovholuknf: passeed --> passed

docusaurus/docs/manage/02-router/01-overview.md
dariuszSki: makes sense, will update it

docusaurus/docs/manage/02-router/01-overview.md
dariuszSki: I guess this falls into the category of personal preference, but I can see it either way, so we can go with your suggestion

docusaurus/docs/manage/02-router/01-overview.md
dariuszSki: makes sense, I always the debate with my self how I should write this as I, we, or it

docusaurus/docs/core-concepts/security/authorization/policies/_create-edge-router-policy-cli.md
dariuszSki: Yes, I noticed, but I thought I got an error once, so was bit confused. It makes sense though if oyu can do that. I guess that relates to the number in front of it as well. For example, 04-cli/05-logging.md, can be as cli/logging

docusaurus/docs/manage/02-router/03-sizing.md
dariuszSki: I changed it to ```Bare Metal```

docusaurus/docusaurus.config.js
dariuszSki: I did that and got the latest file, but the latest file had links to files (router-overview and edge-router) I  merged into one router folder. Thus they are no longer valid. New links need to be created I would think. How is this file generated to update it? I was not able to compile with them being there.

qrkourier: For each broken redirect you need to change the value of `to:` to a valid path. The old value is invalid because you reorganized the pages which now have new paths. For example, `/docs/manage/router/router-deploying-overview` is the new path since you deleted `/docs/manage/edge-router`.

dariuszSki: reread your first comment and it sounds like these are created manually, True?

qrkourier: Correct, the Docusaurus config file is not generated. It's hand crafted.

docusaurus/docs/core-concepts/security/authorization/policies/_create-edge-router-policy-cli.md
dariuszSki: If I remove the file type in the link, I get error `broken links` when I run ```yarn build```

dariuszSki:
```
[ERROR] Unable to build website for locale en.
[ERROR] Error: Docusaurus found broken links!

Please check the pages of your site in the list below, and make sure you don't reference any path that does not exist.
Note: it's possible to ignore broken links with the 'onBrokenLinks' Docusaurus configuration, and let the build pass.

Exhaustive list of all broken links found:

- On source page path = /docs/core-concepts/security/authorization/policies/creating-edge-router-policies:
   -> linking to /docs/manage/04-cli/02-logging-in

- On source page path = /docs/core-concepts/security/authorization/policies/creating-service-edge-router-policies:
   -> linking to /docs/manage/04-cli/02-logging-in

- On source page path = /docs/core-concepts/security/authorization/policies/creating-service-policies:
   -> linking to /docs/manage/04-cli/02-logging-in

- On source page path = /docs/core-concepts/security/authorization/policies/overview:
   -> linking to /docs/manage/04-cli/02-logging-in
```

qrkourier: The correct path is `/docs/manage/cli/cli-login` because you have overridden the category for the `cli` branch of the folder tree.

qrkourier: It's a tradeoff. You could use your original approach of the relative path to the MD(X) file and let Docusaurus resolve it at build time. At least this way it's guaranteed to resolve to a valid link. If we assume the structure of the category is unlikely ever to change then the absolute path to the current, rendered location of the article will be valid even if the filename suffix changes. Personally, I prefer to link to the relative filename and let Docusaurus figure out the HTTP URL path.

docusaurus/docusaurus.config.js
dariuszSki: Ok, after update, we have these:
```
{
            to: '/docs/manage/controller',
            from: ['/ziti/manage/controller.html', '/ziti/manage/manage.html'],
          },
          {
            to: '/docs/manage/router/router-overview',
            from: ['/ziti/manage/router/router-overview.html'],
          },
          {
            to: '/docs/manage/pki',
            from: ['/ziti/manage/pki.html'],
          },
          {
            to: '/docs/manage/troubleshooting/troubleshooting',
            from: ['/ziti/manage/troubleshooting/troubleshooting.html'],
          },
          {
            to: '/docs/manage/cli/cli-overview',
            from: ['/ziti/manage/cli/cli-overview.html'],
          },
```

dariuszSki: Question becomes now, do we need to do this for every file in the router folder or Cli folder. Why do we need this at all?

qrkourier: Certainly not for any new URL paths. These redirects prevent established paths from breaking by handling requests for the old path and redirecting them to a new, valid destination.

docusaurus/docs/core-concepts/security/authorization/policies/_create-edge-router-policy-cli.md
dariuszSki: ok, I see the value of IDs now more clearly. Updated the links to use them instead of file names.

dariuszSki: If you are ok with this, I will close this issue. 
```
[logged in](/docs/manage/cli/cli-login))
[update section](update/router-update/#update-router)
etc
```

qrkourier: That's a valid reference for `logged in`. 

For `update section`, you'll have to make a similar choice between the **absolute URL path** (starts with `/`) and **relative URL path** (starts with `./`, preferably) , and **relative file path** (starts with `./`, preferably, and ends with a filename suffix). The current value doesn't look valid. Does it build like that?

dariuszSki: the relative path works both ways, i.e ./router or just router. Not sure if it makes any difference which one to use.

qrkourier: I use `./` because it's explicit and triggers my IDE to auto-complete the reference.

dariuszSki: ok, since most people use that I assume. Then , it is ```./``` :)

docusaurus/docusaurus.config.js
dariuszSki: ok will update the existing ones to new paths.

docusaurus/docs/manage/02-router/02-configuration.md
dariuszSki: I will remove this for now, we will add the fabric router once the cli option to create transit routers is  added.

docusaurus/docs/manage/02-router/04-update/01-cli.md
dariuszSki: I will remove this for now, we will add the fabric router once the cli option to create transit routers is  added.

qrkourier (approved):
ready to stage the router guide from @dariuszSki in the router overhaul branch

dariuszSki:
@qrkourier we are ready to merge. I don't think I can do it.

PR #246, opened 2022-11-15
Adding a section for create the Edge Router policies

gberl002:
Fixed a typo


docusaurus/docs/quickstarts/network/local-with-docker.md
```diff
@@ -85,20 +85,38 @@ docker run \
   /openziti/scripts/run-controller.sh
 ```
 
+## Create Edge Router Policies
+For the purposes of quickstart, it is necessary to create an Edge Router Policy and a
```
dovholuknf: What do you think of:

OpenZiti requires specific authorization of users and routers (edge router policies), as well as authorization of services and routers  (service edge router policies) . The docker-based quickstart doesn't perform these steps automatically. Run the initialization container one time, after starting the controller as shown

dovholuknf: Oh, and add links to doc pages for each idea

docusaurus/docs/quickstarts/network/local-with-docker.md
```diff
@@ -85,20 +85,39 @@ docker run \
   /openziti/scripts/run-controller.sh
 ```
 
+## Create Edge Router Policies
+OpenZiti requires specific authorization of users and routers (edge router policies), as well as authorization of 
+services and routers (service edge router policies). The docker-based quickstart doesn't perform these steps 
+automatically. Run the initialization container one time, after starting the controller as shown
```
dovholuknf: felt just a tiny bit unfinished to me? maybe just tag on the end "run as shown, to setup the default authorization"?

dovholuknf (approved):

docusaurus/docs-glossary/glossary.md
```diff
@@ -4,12 +4,19 @@ id: glossary
...
+
 ## Network Overlay, Overlay
 A Ziti network is implemented as an "overlay". A network overlay abstracts away the layers beneath it, providing a new set of abstractions for designing and implementing software and systems. Good programming abstractions allow developers to focus on the rules implemented by those abstractions without being concerned with the layers below the abstraction. Ziti's overlay allows developers to focus on connectivity between components without having to be concerned with low-level details of how that connectivity is managed.
 
 ## Service Definition
 A service definition is used to "bind" a service to a specific underlay network expression, through one or more nodes on a Ziti overlay network. A service definition usually includes a terminating router (or routers) and one or more SDK or underlay network endpoints where the service can be reached.
 
+## Service Edge Router Policy
+A Service Edge Router Policy is a policy/configuration that describes which Edge Routers are able to be used for inbound
+or outbound traffic for a particular Service
```
qrkourier, line 18: "inbound and outbound" is a little more clear because "or" implies that the SERP only applies to one or the other for a particular service.

docusaurus/docs-glossary/glossary.md
```diff
@@ -4,12 +4,19 @@ id: glossary
 # Glossary
 Here you will find a list of terms you may come across as you are using Ziti.
 
+## Edge Router Policy
+An Edge Router Policy is a policy/configuration that describes which Identities can connect to which Edge Routers
```
qrkourier, line 8: I prefer to say that identities "use" edge routers because "connect" implies intent when in fact, the intention is to connect to the service, and the identity is using the router to do so. It seems best to sidestep the fact that the SDK is connecting to the router via underlay because we're trying to describe the overlay function of the ERP.

qrkourier (commented):
Review of glossary.md only

docusaurus/docs-glossary/glossary.md
dovholuknf: I dunno, personally I think identities connect to edge routers before connecting to services and this policy is about what routers an identity is permitted to connect to -- to "use". So for me, I feel like it's relevant in this context to say connect.

gberl002:
Branch got messed up due to unsigned commit and my inability to save it, I've created a new branch and new PR here https://github.com/openziti/ziti-doc/pull/326

PR #248, opened 2022-11-20
tweak the hosted quickstart and zac guide

qrkourier:


docusaurus/docs/quickstarts/zac/installation.md
```diff
@@ -33,32 +33,34 @@ you can perform the following steps.
...
 
     ```bash
     cd "${ZITI_HOME}/ziti-console"
     sudo apt install npm nodejs -y
     npm install
     ````
-   
-3. Use the ziti-controller certificates for the Ziti Console:
+
+3. You may need a newer version of Node if the app complains about `await`. In that case, you could use [Node Version Manager](https://github.com/nvm-sh/nvm#readme) (`nvm`) instead of `apt` to install Node.
```
dovholuknf: this should just be expressed in the Prerequisites section.

qrkourier: moved that part up to pre-reqs

docusaurus/docs/quickstarts/network/hosted.md
```diff
@@ -20,25 +20,14 @@ the configuration options available you'll be better equipped to make changes.
 
 The first issue you will need to deal with is opening some ports. A network will consist of at least one controller and 
 at least one edge router. Both of these components will require ports to be open. For the controller you will need to 
-open two ports through your firewall, one port for the REST API, and one the control plane. Edge routers will also 
-require two ports open. One for the links created between routers to form the mesh network, and one port for incoming 
-client connections.
+open a range of ports through your firewall:
 
-The ports you choose are not important but unless you change them these ports will default to the following:
+- `8440/tcp`: Edge Controller listening for clients
```
dovholuknf: 8440 doesn't exist. not sure where that came from. it should be 8441 and should state "Edge Controller API"
6262 is the Control plane, I'd just refer to it as that . it's definitely not 8441

8442 is for Edge Connections or the data plane

docusaurus/docs/quickstarts/network/hosted.md
```diff
@@ -57,12 +46,20 @@ variable. For example:
 export EXTERNAL_DNS="ec2-18-100-100-100.us-east-2.compute.amazonaws.com"
 ```
 
+You may override the default administrator password "admin". The secret value will be stored in the `*.env` file inside the express install home directory so that it can be easily referenced by the `zitiLogin` function.
```
dovholuknf: I'd rather we just change the way expressInstall works to simply generate a password and ask the user if they want to use the generated one or some other password. why not just do that -- and then this doc section isn't needed?

docusaurus/docs/quickstarts/network/hosted.md
dovholuknf: i see you changed this below - i'd rather keep 8441 as the API  use 8440 as the control plane just to minimize the changes

docusaurus/docs/quickstarts/network/hosted.md
```diff
@@ -110,17 +109,15 @@ After the files are generated, you can then install them for use by systemd by r
...
 ```
 
 Now, both the controller and the edge router will restart automatically!  After a few seconds you can then run these
 commands and verify systemd has started the processes and see the status:
 
 ```bash
-sudo systemctl -q status ziti-controller --lines=0 --no-pager
-sudo systemctl -q status ziti-router --lines=0 --no-pager
-
+$ sudo systemctl -q status ziti-controller --lines=0 --no-pager
```
dovholuknf: this makes it ___way___ harder to copy/paste. before it was 'bad enough' because the output is in the same code block but now it's vastly more difficult to copy/paste these commands.

docusaurus/docs/quickstarts/network/hosted.md
```diff
@@ -140,29 +138,78 @@ ubuntu@ip-10-0-0-1:~$ sudo systemctl -q status ziti-router --lines=0 --no-pager
...
+$ sudo ss -lntp | grep ziti
+LISTEN 0      4096               *:8440             *:*    users:(("ziti-controller",pid=5240,fd=10))
+LISTEN 0      4096               *:8441             *:*    users:(("ziti-controller",pid=5240,fd=9)) 
+LISTEN 0      4096               *:8442             *:*    users:(("ziti-router",pid=5272,fd=9))     
+LISTEN 0      4096               *:10080            *:*    users:(("ziti-router",pid=5272,fd=8))     
+```
+
+What are these server ports listening for?
+
+- `8440/tcp`: the Edge Controller listening for API requests. We specified this port as env var `ZITI_EDGE_CONTROLLER_PORT`. You need a firewall exception for this port to allow Ziti identities to obtain an Edge session.
```
dovholuknf: All of this was already lightly discussed above and feels redundant here. it also should just be generically documented anywhere else so that any/every quickstart can reference this same information. i think you should just remove this stuff and either figure out a way to generically document this, or just file an issue to do that work. as is, it'll have to be duplicated on any/every quickstart

docusaurus/docs/quickstarts/network/hosted.md
```diff
@@ -140,29 +138,78 @@ ubuntu@ip-10-0-0-1:~$ sudo systemctl -q status ziti-router --lines=0 --no-pager
...
+$ source ~/.ziti/quickstart/$(hostname)/$(hostname).env
 adding /home/ubuntu/.ziti/quickstart/ip-10-0-0-1/ziti-bin/ziti-v0.20.2 to the path
-ubuntu@ip-10-0-0-1:~$ echo $ZITI_HOME
+
+$ echo $ZITI_HOME
 /home/ubuntu/.ziti/quickstart/ip-10-0-0-1
 ```
 
-## Install Ziti Admin Console (ZAC) [Optional]
+## Install Ziti Admin Console (ZAC)
```
dovholuknf: I don't think these new words add any additional clarity to the words that were there before but I could see a "Next Steps" type of section though. more on that below...

docusaurus/docs/quickstarts/network/hosted.md
```diff
@@ -140,29 +138,78 @@ ubuntu@ip-10-0-0-1:~$ sudo systemctl -q status ziti-router --lines=0 --no-pager
...
+This is an optional server app and web console for Ziti network administration.
+
+[Installation guide](../zac/installation.md)
 
 ## Using the Overlay
 
 Now you have your zero trust overlay network in place, you probably want to try it out. Head on over to
 [the services quickstart](../services) and start the journey to understanding how to use OpenZiti.
+
+## Reset Environment and Begin Again
```
dovholuknf: This whole section destroys the flow that the consumer is 'done' with the quickstart. I think what makes more sense is to have a "Next Steps" section at the very bottom with a few bullets that link to their own page. The next steps that make sense are things like:

* Install ZAC 
* Add a Private Router
* Add a Second Public Router
*  etc. etc. etc
* Delete everything and start over (though I'm not sure why people would actually want to do this)

docusaurus/docs/quickstarts/zac/installation.md
```diff
@@ -13,7 +13,9 @@ and can be accomplished in two basic ways shown below.
 
 It's expected that you're using `bash` for these commands. If you're using Windows we strongly recommend that you install 
 and use Windows Subsystem for Linux (WSL). Other operating systems it's recommended you use `bash` unless you are able to 
-translate to your shell accordingly. 
+translate to your shell accordingly.
+
+You will need `node` and `npm` executables. You may need a newer version of Node than your OS package manager provides. Ubuntu Jammy currently provides Node v12 which is too old. If the app complains about `await` then a newer version is needed. In that case, you could use [Node Version Manager](https://github.com/nvm-sh/nvm#readme) (`nvm`) instead of `apt` to install `node` and `npm`.
```
dovholuknf: i understand the desire to be helpful, but i would just express that node 16+ is needed. go figure out how to install that

qrkourier: > You will need `node` and `npm` executables from Node.js v16+. Ubuntu Jammy currently provides v12.

docusaurus/docs/quickstarts/network/hosted.md
qrkourier: We don't quite have enough for separate pages yet, but I'll go ahead and start organizing it this way.

docusaurus/docs/quickstarts/network/hosted.md
qrkourier: - `8440/tcp`: Edge Controller providing router control plane
- `8441/tcp`: Edge Controller providing client sessions
- `8442/tcp`: Edge Router providing client connections
- `8443/tcp`: Ziti Admin Console (ZAC) [optional]

docusaurus/docs/quickstarts/network/hosted.md
qrkourier: I was thinking go ahead and offer them the chance to set a preferred password even after we start generating a random pw. We could just leave it at "change it later if you want" now that's been added to "Next Steps".

docusaurus/docs/quickstarts/network/hosted/index.md
```diff
@@ -1,168 +1,218 @@
...
+LISTEN 0      4096               *:8440             *:*    users:(("ziti-controller",pid=5240,fd=10))
+LISTEN 0      4096               *:8441             *:*    users:(("ziti-controller",pid=5240,fd=9)) 
+LISTEN 0      4096               *:8442             *:*    users:(("ziti-router",pid=5272,fd=9))     
+LISTEN 0      4096               *:10080            *:*    users:(("ziti-router",pid=5272,fd=8))     
+```
+
+## Adding Environment Variables Back to the Shell
+
+If you log out and log back in again you can source the '.env' file located at:
+`source ~/.ziti/quickstart/$(hostname -s)/$(hostname -s).env`.
```
qrkourier: See complementary change in https://github.com/openziti/ziti/pull/909

docusaurus/docs/quickstarts/network/hosted.md
qrkourier: > A random password for user "admin" will be generated and stored in the `*.env` file inside the express install home directory so that it can be easily referenced by the `zitiLogin` function. Optionally, you may override with a preferred password by setting `ZITI_PWD` now.

docusaurus/docs/quickstarts/network/hosted.md
qrkourier: fixed with separate "run me" blocks and "example output" blocks

docusaurus/docs/quickstarts/network/hosted.md
qrkourier: removed

docusaurus/docs/quickstarts/network/hosted/index.md
```diff
@@ -1,168 +1,258 @@
...
-Once you have the network up and running, if you want to install the UI management console, the ZAC, [follow along with
-the installation guide](../zac/installation)
-
-## Using the Overlay
-
-Now you have your zero trust overlay network in place, you probably want to try it out. Head on over to
-[the services quickstart](../services) and start the journey to understanding how to use OpenZiti.
+---
+sidebar_position: 10
+---
```
dovholuknf: why move it just to move it? is there any reason/rationale?

docusaurus/docs/quickstarts/network/hosted/index.md
```diff
@@ -1,168 +1,258 @@
...
+IP addresses in case your IP changes.
+
+The quickest and easiest thing to do, is simply find your external DNS name and set it into the EXTERNAL_DNS environment
+variable. For example:
+
+```bash
+export EXTERNAL_DNS="acme.example.com"
+```
+
+A random password for user "admin" will be generated and stored in the `*.env` file inside the express install home directory so that it can be easily referenced by the `zitiLogin` function. Optionally, you may override with a preferred password by setting `ZITI_PWD` now.
```
dovholuknf: i'm quite torn on "non-line-breaking" markdown like this. part of me likes it because moving text is easier to see, part of me dislikes it because it's not very "markdown-y". Have you found any good guidance on "the right way" to markdown?  I've been wrapping at around 120 chars (not hard, just ish) as a good page width. i'm open to NOT doing that, i would just like to be consistent.  also reviewing the doc here on GH is easier when it's wrapped ~120 ish chars

gberl002: I have also been hard wrapping at 120 for two reasons, one is that it was already being done, two is because I find it easier to read locally without needing to use a MD renderer to soft wrap.

Now that I've said that "out loud" I have to believe there is a soft wrap option in GoLand that can be set to MD only.

docusaurus/docs/quickstarts/network/hosted/index.md
```diff
@@ -1,168 +1,258 @@
...
+export ZITI_EDGE_CONTROLLER_IP_OVERRIDE="${EXTERNAL_IP}"
+export ZITI_EDGE_ROUTER_IP_OVERRIDE="${EXTERNAL_IP}"
+export ZITI_EDGE_CONTROLLER_HOSTNAME="${EXTERNAL_DNS}"
+export ZITI_EDGE_ROUTER_HOSTNAME="${EXTERNAL_DNS}"
+export ZITI_CTRL_PORT=8440
+export ZITI_EDGE_CONTROLLER_PORT=8441
+export ZITI_EDGE_ROUTER_PORT=8442
+
+# now download, source, and execute the expressInstall function
+source <(wget -qO- https://raw.githubusercontent.com/openziti/ziti/main/quickstart/docker/image/ziti-cli-functions.sh)
```
gberl002: Isn't this going to break for Bash 3 again? I thought that was the reason for `source /dev/stdin <<< ...`

dovholuknf: good catch, geoff. definitely don't want to change the way that gets sourced. on mac's the bash is ancient and this was necessary

docusaurus/docs/quickstarts/network/hosted/index.md
```diff
@@ -1,168 +1,258 @@
...
+Inspect the servers' listening ports.
+
+```bash
+# run me
+sudo ss -lntp | grep ziti
+```
+
+```bash
+# example
+$ sudo ss -lntp | grep ziti
```
gberl002: I don't have `ss` available on Mac, I'm not sure if it's standard in Linux. In the past, I've used `lsof` which does ship with Mac but I'm also not sure if that is standard on Linux.

docusaurus/docs/quickstarts/network/hosted/index.md
gberl002: And in fact there is... View-->Active Editor-->Soft-Wrap. It only applies to the currently active editor so, not all MD files but I would be fine with this workaround if we decided to go without hard line breaks.

docusaurus/docs/quickstarts/network/hosted/index.md
qrkourier: Because you'd suggested separate pages for the next steps, but there's not quite enough content yet to break those out into pages. I'd already moved hosted.md to hosted/index.md by the time I realized that, and expect it'll need to be moved anyway to facilitate created those forthcoming child pages for next steps.

docusaurus/docs/quickstarts/network/hosted/index.md
qrkourier: Glad you found that. I'm using soft wrap in my editor so hard wraps are of no immediate value to me or most readers. Let's soft wrap in our editors to ease reading and editing and optimize for the following in commits:

1. vast majority of readers are reading styled HTML, not Markdown, so hard wraps have no value to them
2. minimize visible differences in Git diff i.e. if we hard re-wrap after changing a line that creates a lot of visual noise in the diff.

docusaurus/docs/quickstarts/network/hosted/index.md
qrkourier: I too have found the `/dev/stdin` approach was necessary for default BASH 3 that shipped with latest macOS in 2013. Haven't checked since then. I guessed wrong that this was aimed at Linux servers. I'll have to redress a lot of assumptions, e.g. portable vs modern BASHisms. This is pretty bad news for the script.

qrkourier: Zsh is the default since macOS Catalina (2019) and so most macOS users will be able to use modern shell-isms.

qrkourier: We need a preamble near the top of this quickstart like "This is intended for Linux servers" or "This quickstart should work with modern macOS or Linux or WSL". I think this particular hosted-anywhere quickstart lends itself to the Linux server audience and I imagine the WSL and macOS crowds would use one of the local quickstarts instead.

docusaurus/docs/quickstarts/network/hosted/index.md
qrkourier: Do you think we should say the hosted anywhere quickstart works on macOS? Seems like the local quickstart is a better fit for workstations. It makes sense for this one to be aimed squarely at Linux servers because of the need to open firewall ports and such, the intention to make the controller and router publicly reachable.

docusaurus/docs/quickstarts/network/hosted.md
```diff
@@ -1,168 +1,235 @@
...
+Controller systemd file written to: /home/ubuntu/.ziti/quickstart/ip-172-31-23-18/ip-172-31-23-18-edge-controller.service
+
+$ createRouterSystemdFile "${ZITI_EDGE_ROUTER_RAWNAME}"
+Router systemd file written to: /home/ubuntu/.ziti/quickstart/ip-172-31-23-18/ip-172-31-23-18-edge-router.service
+```
+
+After the files are generated, you can then install them for use by systemd by running:
+
+```bash
+# run me
```
dovholuknf: i never see comments like this on other code snippets fwiw. i think it's pretty clear that it's meant to be run. i don't see value in these "run me" comments

docusaurus/docs/quickstarts/network/hosted.md
```diff
@@ -1,168 +1,235 @@
...
+sudo cp "${ZITI_HOME}/${ZITI_EDGE_CONTROLLER_RAWNAME}.service" /etc/systemd/system/ziti-controller.service
+sudo cp "${ZITI_HOME}/${ZITI_EDGE_ROUTER_RAWNAME}.service" /etc/systemd/system/ziti-router.service
+sudo systemctl daemon-reload
+sudo systemctl enable --now ziti-controller
+sudo systemctl enable --now ziti-router
+```
+
+```bash
+# example output
```
dovholuknf: I'd just output the actual example output, this is not acurate. i'd also title "example output" outside of the bash script for even further clarity. i do like separating the 'actual output' though. that's been a stick in my craw for a while

docusaurus/docs/quickstarts/network/hosted.md
```diff
@@ -1,168 +1,235 @@
...
+
+### Install Ziti Admin Console (ZAC)
+
+This is an optional server app and web console for Ziti network administration.
+
+[Installation guide](../zac/installation.md)
+
+### Add a Second Public Router
+
+Routers automatically form a mesh of router links when there's more than one. In order for your express-installed router to participate in the mesh it needs a firewall exception to expose its "link listener" on port `10080/tcp`, the default port.
```
dovholuknf: > Routers automatically form a mesh of router links when there's more than one.

this is not factually true and is a complication that doesn't need to be on this page. there's no need for exposition here. i thought we were going to go with a "next steps" idea for down here?

docusaurus/docs/quickstarts/network/hosted.md
```diff
@@ -1,168 +1,235 @@
...
+
+<!-- TODO: link to the new router deployment guide when it's published -->
+
+### Change Admin Password
```
dovholuknf: this should be a linked page and a 'next step' imo

docusaurus/docs/quickstarts/zac/installation.md
```diff
@@ -1,26 +1,24 @@
...
-
-### Prerequisites
+## Prerequisites
 
 It's expected that you're using `bash` for these commands. If you're using Windows we strongly recommend that you install 
 and use Windows Subsystem for Linux (WSL). Other operating systems it's recommended you use `bash` unless you are able to 
-translate to your shell accordingly. 
+translate to your shell accordingly.
+
+You will need `node` and `npm` executables from Node.js v16+. Ubuntu Jammy currently provides v12.
```
dovholuknf: stating Jammy provides v12 makes me wonder about fedora, or other operating systems. it's better to just say "get the right version for your OS" and leave it at that.

docusaurus/docs/quickstarts/zac/installation.md
```diff
@@ -1,26 +1,24 @@
...
 
 :::note
 When running Ziti Administration Console, you should also prefer using https over http. In order to do this you will need
 to either create, or copy the certificates needed. Each section below tries to show you how to accomplish this on your own.
 :::
 
-### Cloning From GitHub
+## Run ZAC by Cloning From GitHub
```
dovholuknf: this is an 'installation' markdown. `Run ZAC by` is redundant. The previous titles were better imo

docusaurus/docs/quickstarts/zac/installation.md
```diff
@@ -33,32 +31,31 @@ you can perform the following steps.
...
 
     ```bash
     cd "${ZITI_HOME}/ziti-console"
-    sudo apt install npm nodejs -y
     npm install
     ````
-   
-3. Use the ziti-controller certificates for the Ziti Console:
+
+1. Use the ziti-controller certificates for the Ziti Console:
```
dovholuknf, line 42: this could probably be enhanced to just talk about how the certs need to be configured for ZAC and then state that you can just use the same certs from the controller, but they browser won't automatically trust them. i think that's a whole separate page that might be useful though, and not one that I would stuff all in here. Mabe just a "see here how to configure ZAC for TLS" or something like taht?

docusaurus/docs/quickstarts/zac/installation.md
```diff
@@ -181,10 +181,14 @@ internal docker compose network that is declared in the compose file. You would
...
-3. Login with admin/admin
- 
+3. Login as admin
+
+   If you followed the quickstart then admin's password is stored as an environment variable `ZITI_PWD` defined in your quickstart environment file. You may reload your quickstart environment by sourcing the environment file: `source ~/.ziti/quickstart/$(hostname -s)/$(hostname -s).env`.
+
    ![img_2.png](./zac_login.png)
 
-4. **IMPORTANT!!!** Edit your profile and change the password
+4. Optionally, edit your profile to change admin's password
```
dovholuknf: i don't think this whole section is necessary now given the other changes to the quickstart. the password will already be randomized

docusaurus/docs/quickstarts/zac/installation.md
```diff
@@ -181,10 +181,14 @@ internal docker compose network that is declared in the compose file. You would
...
 
    ![img_3.png](./zac_change_pwd.png)
+
+   Keep in mind that this is the same password that is sourced from the quickstart environment file. You'll need to change it there too for the `zitiLogin` command.
```
dovholuknf: same comment from before - don't need it. also redundant for what is in step 3. maybe just combine some of this with step 3 and eliminate 4 entirely?

dovholuknf:
enough people use the quickstarts where i think we should also include changes like this in the changelog for the /ziti project as well. i don't recall if that was done (i could check but writing this comment was easier)

docusaurus/docs/quickstarts/network/hosted.md
```diff
@@ -1,168 +1,235 @@
...
+configuration files that are output.  The script allows you to configure these values and if you are deploying to a cloud
+provider (AWS, Azure, OCI, IBM, Digital Ocean, GCP, etc.) you will almost certainly want to use the **public** DNS name
+of your instance. You possibly  want to override the IP address to use as well, but using DNS names is superior to using
+IP addresses in case your IP changes.
+
+The quickest and easiest thing to do, is simply find your external DNS name and set it into the EXTERNAL_DNS environment
+variable. For example:
+
+```bash
+export EXTERNAL_DNS="acme.example.com"
```
dovholuknf: @gberl002 what happens when there IS no external dns? i think you recently changed these words? or are you in that procecss? (i don't recall)

docusaurus/docs/quickstarts/network/hosted.md
```diff
@@ -1,168 +1,235 @@
...
+export ZITI_EDGE_CONTROLLER_HOSTNAME="${EXTERNAL_DNS}"
+export ZITI_EDGE_ROUTER_HOSTNAME="${EXTERNAL_DNS}"
+export ZITI_CTRL_PORT=8440
+export ZITI_EDGE_CONTROLLER_PORT=8441
+export ZITI_EDGE_ROUTER_PORT=8442
+```
+
+```bash
+# run expressInstall
+source /dev/stdin <<< "$(wget -qO- https://raw.githubusercontent.com/openziti/ziti/main/quickstart/docker/image/ziti-cli-functions.sh)"; expressInstall
```
dovholuknf, line 70: i'd be ok with slapping this file into this size and shrinking the "get ziti" command if you want too?

dovholuknf: oh - or just a redirect from this site to the 'big long url' and instruct wget to follow redirects?

docusaurus/docs/quickstarts/network/hosted.md
qrkourier: This is in the "next steps" section.

> In order for multiple routers to form transit links, they need a firewall exception to expose the "link listener" port. The default port is `10080/tcp`.

docusaurus/docs/quickstarts/network/hosted.md
qrkourier: There doesn't seem to be enough content in the next steps section to justify separate pages, and we're currently in this state after first creating separate pages and then reverting to a single page. I planned to break these next steps sections out into separate pages if/when they are longer.

dovholuknf: I very much disagree. 
1. i don't want to see this on the page - it's just extra noise
2. we're already addressing this issue by generating the admin password now, so this section is probably not even needed now.
3. there are 3 other quickstart documents. are you planning to copy/paste this to all of those? it'll be so much easier to just have a "how to change your password" page and reference it. we need that anyway, as it is.

docusaurus/docs/quickstarts/zac/installation.md
qrkourier: Sounds like the beginnings of a ZAC deployment guide for the new Ops section.

docusaurus/docs/quickstarts/network/hosted.md
```diff
@@ -1,168 +1,235 @@
...
+$ zitiLogin
+Token: d6152c84-3166-4ae4-8ca3-1c38c973d450
+Saving identity 'default' to /home/ubuntu/.ziti/quickstart/ip-172-31-28-116/ziti-cli.json
+
+$ ziti edge update authenticator updb -s
+Enter your current password: 
+Enter your new password: 
+```
+
+### Delete everything and start over
```
dovholuknf: This is also 'noise' and not relevant to an actual user too. That's why I think this fits as a different page/link. it also uses a command I find 'scary': `rm -r "${ZITI_HOME}"` that can be messed up pretty easilyimo. the unsetZitiEnv  could use a whole lot more words too. it just doesn't add clarity to the quickstarts and is better as a reference

docusaurus/docs/quickstarts/network/hosted.md
qrkourier: I've gotten into this habit when I intend to run the hosted quickstart but don't plan to expose it externally.

```bash
export EXTERNAL_DNS="localhost"
```

That way, I don't need to modify the provided shell env snippet which currently needs `EXTERNAL_DNS`. We could provide a default value for `EXTERNAL_DNS` in that snippet. I'll add that in this branch for review.

gberl002: If `EXTERNAL_DNS` isn't set then the controller advertised address would resort to `ZITI_EDGE_CTRL_ADVERTISED_HOST_PORT `which should default to `"${ZITI_EDGE_CONTROLLER_HOSTNAME}:${ZITI_EDGE_CONTROLLER_PORT}"`

Also, `EXTERNAL_DNS` gets added to the PKI allow list if it is set.

gberl002: The `web->bind->address` and `edge->api->address` are set using `EXTERNAL_DNS` if it's set, otherwise they'll be  `ZITI_EDGE_CTRL_ADVERTISED_HOST_PORT`

docusaurus/docs/quickstarts/network/hosted.md
qrkourier: Redirect added in https://github.com/openziti/ziti-doc/pull/265

docusaurus/docs/quickstarts/network/hosted.md
qrkourier: I've added some words. See what you think. It feels important to maintain a procedure for starting over for reasons like: wanting to re-run with different params, something broken and user is troubleshooting.

docusaurus/docs/quickstarts/network/hosted.md
qrkourier: I think this means we should default `EXTERNAL_DNS` to empty string `""`, not `localhost`, in this hosted quickstart's `expressInstall` setup section. For example:

```bash
export EXTERNAL_DNS=${EXTERNAL_DNS:-""}
export EXTERNAL_IP="$(curl -s eth0.me)"       
export ZITI_EDGE_CONTROLLER_IP_OVERRIDE="${EXTERNAL_IP}"
export ZITI_EDGE_ROUTER_IP_OVERRIDE="${EXTERNAL_IP}"
export ZITI_EDGE_CONTROLLER_HOSTNAME="${EXTERNAL_DNS}"
export ZITI_EDGE_ROUTER_HOSTNAME="${EXTERNAL_DNS}"
export ZITI_CTRL_PORT=8440
export ZITI_EDGE_CONTROLLER_PORT=8441
export ZITI_EDGE_ROUTER_PORT=8442
```

That way, if they don't set `EXTERNAL_DNS` it'll default to `ZITI_EDGE_CTRL_ADVERTISED_HOST_PORT`.

qrkourier: Actually, an unset var is practically equivalent to an empty string value in this particular case because we're not catching errors (`errexit`) or unset vars (`nounset`) because the shell is presumed interactive. So, we don't really need to add the line to explicitly set `EXTERNAL_DNS`. If they skip the setup step to assign `EXTERNAL_DNS` then it'll just work.

docusaurus/docs/quickstarts/network/help/_category_.json
```diff
@@ -0,0 +1,4 @@
+{
```
dovholuknf, line 1: What do you think about putting these under 'guides' or at the top level? I kinda like help back up at the top level but I'm not sure how many 'help' things we'll end up with. seems like it might be more useful up there?

qrkourier: They're specific to the network and console quickstarts.

docusaurus/docs/quickstarts/network/help/change-admin-password.md
```diff
@@ -0,0 +1,25 @@
+---
+title: Change Password
+---
+
+Follow these instructions to change your quickstart admin's password. A random password was generated for you when you ran `expressInstall` and saved in your *.env file as `ZITI_PWD`. After changing the password, also change the value of `ZITI_PWD` in `~/.ziti/quickstart/$(hostname -s)/$(hostname -s).env` to match. This variable is used by the `zitiLogin` function.
+
+## Use the `ziti` CLI {#ziti-cli}
+
+```bash
+$ zitiLogin
```
dovholuknf, line 10: zitiLogin is helpful. do you think we should keep it or start using `ziti edge login` instead? i feel like this whole page should avoid mentioning the quickstart stuff regardless since these are totally ancillary to the quickstart.

docusaurus/docs/quickstarts/network/help/change-admin-password.md
```diff
@@ -0,0 +1,25 @@
...
+$ ziti edge update authenticator updb -s
+Enter your current password: 
+Enter your new password: 
+```
+
+## Use the Ziti Console {#ziti-console}
+
+If you installed the Ziti Console (ZAC) in your quickstart network then you can [login with your current password](../../../quickstarts//zac/installation.md#login-and-use-zac) and change it.
+
+Navigate to your user profile and change the password as pictured.
```
dovholuknf: I always try to be more clear and help the user find what they are clicking on since screen caps don't do a good job of capturing mouse position. To me It's more clear to indicate that the user should:

> click on the person icon on the lower left and then select "Edit Profile" as shown to change your password

dovholuknf (approved):

docusaurus/docs/quickstarts/network/help/change-admin-password.md
qrkourier: It does feel like an opportunity to introduce the `ziti` CLI, but the login command is pretty long with the certificate param.

```bash
$ type zitiLogin

zitiLogin is a function
zitiLogin () 
{ 
    "${ZITI_BIN_DIR-}/ziti" edge login "${ZITI_EDGE_CTRL_ADVERTISED}" -u "${ZITI_USER-}" -p "${ZITI_PWD}" -c "${ZITI_PKI_OS_SPECIFIC}/${ZITI_EDGE_CONTROLLER_ROOTCA_NAME}/certs/${ZITI_EDGE_CONTROLLER_INTERMEDIATE_NAME}.cert"
}
```

Even the console stuff is quickstart-specific unless we assume they already know how to login to the console. If we trim the quickstart-specific stuff all that remains is the `ziti edge update authenticator updb -s` which I was going to add to the CLI Basics page anyway.

qrkourier: added the change updb hint in https://ziti-doc-git-issue-106-router-docs-overhaul-openziti.vercel.app/operations/cli-basics#change-password

PR #260, opened 2022-12-06
update blog article to use the preferred Helm chart

qrkourier:
…part

### DIFF
--- a/docusaurus/blog/zitification/kubernetes/kubernetes-oci.md
+++ b/docusaurus/blog/zitification/kubernetes/kubernetes-oci.md
@@ -2,10 +2,6 @@
 authors: dovholuknf
 ---
 
-<head>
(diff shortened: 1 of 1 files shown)


qrkourier:
@dovholuknf We don't yet have a solution for syndicating canonical URLs from HashNode. I updated this article in the Docusaurus blog (which is still hidden in navigation) because it contained references to a deprecated Helm chart. It's nearly identical to the HashNode counterpart article which is the canonical article, and so I added the canonical URL according to Docusaurus page metadata instructions, which don't seem to be working. I have a thread in their Discord about that, but no solution to that either, yet.

qrkourier:
@dovholuknf this PR updates a blog post to use the preferred Helm chart. I intend to drain users away from the non-preferred chart.

dovholuknf (approved):

PR #257, opened 2022-11-30
Issue 106 router docs overhaul

qrkourier:

PR #264, opened 2022-12-07
repair broken links

### LINKED ISSUE #263: detect broken links

qrkourier:
~~I'll set up the broken link detector job in the GH pages repo so it can run after a successful deployment.~~

This draft PR is based on #257 so it needs to wait until that merges before marking the PR ready for review.


qrkourier:
The broken link checker is set up to run on the same schedule as the `publish.sh` job. It's configured and working, and the broken links it's currently detecting are fixed in this branch.

docusaurus/docs-guides/hsm/softhsm.mdx
```diff
@@ -1,1301 +1,1301 @@
-import Tabs from '@theme/Tabs';
```
dovholuknf, line 1: why are there so many changes on this one file?

qrkourier: There's something going on with line endings, though the file wasn't moved or line endings intentionally rewritten. These are the actual changes.

```
diff --git a/docusaurus/docs-guides/hsm/softhsm.mdx b/docusaurus/docs-guides/hsm/softhsm.mdx
index 3e837ee..6c40698 100644
--- a/docusaurus/docs-guides/hsm/softhsm.mdx
+++ b/docusaurus/docs-guides/hsm/softhsm.mdx
@@ -330,13 +330,13 @@ Ensure you use the correct dll. If you use an x86 dll with x64 binaries you'll g
 
 ### Putting It All Together
 
-Above we've only shown the commands that need to run and not what the output of those commands would look like. Here
-we'll see all the commands put together along with all the output from the commands. This section is long - you are
-warned! Also note that this content is subject to change. If the output you see is not identical it's because the
+Above, we've only shown the commands that need to run and not what the output of those commands would look like. Here
+we'll see all the commands put together, along with all the output from the commands. This section is long - you are
+warned! Also, note that this content is subject to change. If the output you see is not identical it's because the
 software has changed since this information was captured. File an [issue](https://github.com/openziti/ziti-doc/issues)
 if you'd like to see it updated.
 
-# [Sample Output](#tab/hidden-linux)
+## Sample Output
 
 The tabs to the right contain example output from running all the commands in sequence. If you want to see what the
 output would likely look like click one of the tabs to the right. Reminder - it's a lot of commands and a lot of output!
```

qrkourier: The _real_ differences are highlighted but everything is shaded so they're really inconvenient to see.

qrkourier:
```bash
git diff --ignore-cr-at-eol main -- docusaurus/docs-guides/hsm/softhsm.mdx
```

dovholuknf: Can you please redo the PR with just the link changes and not everything else?

qrkourier: Yep. https://github.com/openziti/ziti-doc/pull/279

PR #267, opened 2022-12-07
remove an unnecessary click to open the zac quickstart

qrkourier:


dovholuknf:
the title of the pr is insufficient alone for me to understand what you're doing here and why. also it has some kind of precommit hook built into it too. please discuss this sort of change before implementing it. remove the hook and action and the rest looks ok

.pre-commit-config.yaml
```diff
@@ -1,3 +1,4 @@
+repos:
```
qrkourier, line 1: > the title of the pr is insufficient alone for me to understand what you're doing here and why. also it has some kind of precommit hook built into it too. please discuss this sort of change before implementing it. remove the hook and action and the rest looks ok

@dovholuknf I think the pre-commit config was unexpected because I'd copied it from the branch where it had merged to main yesterday in a parallel pull so I could also work in this branch. Is it good now that the merge conflict is resolved?

dovholuknf: the whole `.pre-commit-config.yaml` should be removed. it's still in the branch

dovholuknf (approved):

PR #276, opened 2022-12-08
add a guide for configuring alt server certs across Ziti components

qrkourier:


dovholuknf:
can you add a link back to this guide from the feature request/bug that covers making the identity.ca and identity.cert fields are not required? that was we update this guide when that gets fixed?

dovholuknf (approved):

PR #277, opened 2022-12-08
lint zac quickstart

qrkourier:


dovholuknf (approved):
it's fine. i'd prefer not to see PRs that are like this though if we can avoid them. approving this one

PR #293, opened 2022-12-20
dry the proxy script and add deploy notes

qrkourier:


docusaurus/docs/core-concepts/clients/tunnelers/linux/index.md
```diff
@@ -42,7 +42,7 @@ Architectures available:
 * arm64
 
 ```bash
-curl -sSLf https://get.openziti.io/pack/package-repos.gpg \
+curl -sSLf https://get.openziti.io/tun/package-repos.gpg \
```
dovholuknf, line 45: not sure i understand this change. pack seemed fine. if anything, why not get rid of this entirely. would this file eventually have all the needed material for ziti-edge-tunnel AND ziti AND (etc.)?

qrkourier: I decided on the more generic "tun" abbreviation for tunneler-related links. "Pack" seemed fine for package-related links, but there's no reason to maintain each when "tun" will suffice for both.

qrkourier: To your question about the file `package-repos.gpg`. There is a misunderstanding. That file contains the package signing pubkey that package users should trust. The `get.openziti.io` URL-shortener merely abbreviates URL paths for `raw.githubusercontent.com`, and doesn't know about particular files.

dovholuknf (approved):

PR #296, opened 2022-12-21
Copy-check yubikey guide

qrkourier:


dovholuknf:
i'm going to park this issue for now since the whole article needs to be redone with `ziti-edge-tunnel` along with a potential update for ZDEW/linux (possibly mac if it can work, gonna ask to find out). so this PR will serve as a reminder that it needs to be updated.

PR #300, opened 2023-01-03
raise a GH issue if broken links

qrkourier:
And close the issue if links are not broken


dovholuknf (approved):
seems like it'd be better to get a notification on a failing job, instead of making an issue.  `¯\_(ツ)_/¯ `

PR #304, opened 2023-01-04
clarify  use of alt_server_certs

qrkourier:


docusaurus/docs-ops/30-configuration/conventions.md
```diff
@@ -46,8 +46,7 @@ db: ${ZITI_DATA}/db/ctrl.db
 OpenZiti uses a common framework for loading, storing, and processing certificate and private key configuration.
 Identity sections all have a similar format. The use of the defined certificates is up to the implementing application.
 So see their configuration sections for details on which values are utilized for what. This documentation provides an
-overview useful to understand the "default" assumptions. These sections can be formatted as YAML (as is the case for the
-router and controller configurations) or as JSON.
+overview useful to understand the "default" assumptions. These sections can be formatted as YAML. JSON is a subset of YAML.
```
andrewpmartinez: I don't understand the statement "JSON is a subset of YAML". If we are talking about the formats themselves, not true. If we are talking about the identity configuration blocks, they are equivalent between formats.

andrewpmartinez (changes requested):
See YAML/JSON comment.

docusaurus/docs-ops/30-configuration/conventions.md
qrkourier: added clarification

andrewpmartinez (approved):

PR #317, opened 2023-01-05
fix the job run link in the broken links issue template

qrkourier:


.github/workflows/check-links.yml
```diff
@@ -40,3 +40,4 @@ jobs:
           filename: .github/workflows/check-links-issue-template.md
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
```
dovholuknf, line 43: is there some bigger picture i'm missing here? did this not work as it was? this doesn't seem like it's very useful to me if it used to work🤷

qrkourier: It was broken ([issue with broken link](https://github.com/openziti/ziti-doc/issues/311)). The template may use the GH context but must use mustache syntax, not Actions syntax. The easiest way to define pass the URL value is with an env var because referencing the GH context via the webhook payload is more obtuse and verbose.

dovholuknf (approved):

PR #321, opened 2023-01-06
Chore/fabric reference

sabedevops:
The link proxy needs to be updated before this can be merged @qrkourier


qrkourier:
I'm not ready to add a Fabric API reference. Let's revisit in a few weeks.

PR #333, opened 2023-01-10
redirect popular blog links

qrkourier:
change default origin in openziti.io proxy to docs site while handling popular blog links as redirects to blog site


dovholuknf:
is there a way to be able to test this sort of change somehow?

qrkourier:
Sanity checked with CloudFront's built-in functional tester. That eliminates the majority of the risk. I can show you how it works if you want. I didn't promote this new version to Production yet.

dovholuknf:
> Sanity checked with CloudFront's built-in functional tester. 

Is that a step that should be added to doc somewhere? I only asked because it's not possible for me to do it myself "easily" is all

dovholuknf (approved):

PR #376, opened 2023-01-25
Fix CSR

ckdarby:
CSR did not have the "signing" and I first cased each letter of CSR

### DIFF
--- a/docusaurus/docs/learn/introduction/_components.md
+++ b/docusaurus/docs/learn/introduction/_components.md
@@ -15,7 +15,7 @@ The OpenZiti Controller must be configured with public key infrastructure
 (PKI). The configured PKI is used to create secure, mutually-
 authenticated TLS (mTLS) network connections between any two
 pieces of the OpenZiti Network. The OpenZiti Controller does not provide its
-own PKI but for the OpenZiti Controller to sign Certificate Signing Request (CSR)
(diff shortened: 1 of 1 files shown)


docusaurus/docs/learn/introduction/_components.md
```diff
@@ -15,7 +15,7 @@ The OpenZiti Controller must be configured with public key infrastructure
 (PKI). The configured PKI is used to create secure, mutually-
 authenticated TLS (mTLS) network connections between any two
 pieces of the OpenZiti Network. The OpenZiti Controller does not provide its
-own PKI but for the OpenZiti Controller to sign certificate requests (CSR)
+own PKI but for the OpenZiti Controller to sign Certificate Signing Request (CSR)
```
dovholuknf: thanks. I think on this one it probably should stay lower-case and plurarize 'request'. I checked wikipedia and they don't propercase "certificate signing request".

dovholuknf: or - "to sign a certificate signing request". That makes sense too

ckdarby: :+1:

ckdarby:
@dovholuknf All good, lower cased as suggested.

dovholuknf (approved):
thanks again!

PR #378, opened 2023-01-26
Fix guides index

qrkourier:


docusaurus/sidebars.js
```diff
@@ -16,7 +16,7 @@ const sidebars = {
   docsSidebar: [
     {
       type: 'html',
-      value: 'LEARN', // The HTML to be rendered
+      value: '<a href="/docs/learn/introduction/">LEARN', // The HTML to be rendered
```
dovholuknf: this is improper html. this 'works'?

dovholuknf: If you're going with the `<a>` approach -- i dunno if we should, you need to style the link to look the same. notice the REFERENCE difference

image

qrkourier: Forgot the closing tag.

qrkourier:
@dovholuknf this PR fixes the sidebar heading links and the misplaced index page for the Guides section

dovholuknf (dismissed):

dovholuknf (approved):

PR #405, opened 2023-02-14
clarify router listeners

qrkourier:


docusaurus/docs/reference/30-configuration/router.md
```diff
@@ -289,18 +289,16 @@ link:
 ### `listeners`
 
 Listeners configure different types of server logic and protocols to be "listened" for on the router. This includes
-opening ports on one or more interfaces by specified in the `address` option. The `address` field is formatted according
-to the [address conventions](conventions.md#addressing). If necessary to refer to the listener externally in
-documents or payloads maintained outside the process, the `advertise` option will be used to format URLs or any
-other address. The `advertise` value should be externally routable.
+opening ports on one or more interfaces specified in the `address` option. The `address` field is formatted according
+to the [address conventions](./conventions.md#addressing). The `advertise` option is used to format the listener address when it is presented to audiences outside the router process. The `advertise` value must be addressable by clients of the listener.
```
dovholuknf: > The `advertise` option is used to format the listener address

I don't think that's actually accurate, right? it's not used to format the address, it **IS** the address.

I prefer this:
> The `advertise` field specifies the address a client will use to connect to the listener. 

shorter, sweeter, still the same content

qrkourier:
Ready for review @dovholuknf

dovholuknf (approved):

PR #455, opened 2023-03-23
Create open ziti network using digital Ocean cloud

om-veer:
Created the step by step to Create open ziti network using digital Ocean cloud. There are some png file which use to create the diagram or screenshot of the result.


dovholuknf:
These files are not part of the docusaurus site and are at the root. We can't merge this as is.

PR #464, opened 2023-03-31
submit digital ocean guide

om-veer:


dovholuknf:
The image folder is misspelled. Also for static assets we probably should consider moving them into a folder under `docusaurus/static/img/`

docusaurus/docs/guides/openziti_Digital_Ocean/Controller_Config.md
```diff
@@ -0,0 +1,248 @@
...
+- 8440/tcp: Edge Controller providing router control plane
+- 8441/tcp: Edge Controller providing client sessions
+- 8442/tcp: Edge Router providing client connections
+- *8443/tcp: Ziti Admin Console (ZAC) [optional]*
+
+## 1.4 Prerequisites
+**Make sure you have jq and curl installed**
+```
+export EXTERNAL_DNS="$(curl -s eth0.me)"
+export EXTERNAL_IP="$(curl -s eth0.me)"
```
smilindave26, line 30: Both EXTERNAL_DNS and EXTERNAL_IP are set to the same value. Is that intentional?

om-veer: Just to keep simple for begeners. We can use network name instead of ext IP. I'll modify the document according after addition of more feature.

om-veer:
Putting some more input as suggested by team.

PR #492, opened 2023-04-25
document tunnel env vars

qrkourier:


qrkourier:
I plan to update this with a link to CLANG reference when it becomes available.

ekoby (approved):

docusaurus/docs/reference/tunnelers/linux/linux-tunnel-options.md
```diff
@@ -1,6 +1,17 @@
 
 # Options and Modes
 
+## `ziti-edge-tunnel` Environment Variables
```
dovholuknf, line 4: This fits better under troubleshooting -- and already has `--verbose 4` listed there. I also don't like documenting things like this since it feels incomplete. if we want to expose these from the tunneler, imo, the tunneler should be exposing these things in a 'tunneler' way. instead of documenting the interim step we should put an issue in to "do this correctly" via the tunneler itself

qrkourier: https://github.com/openziti/ziti-tunnel-sdk-c/issues/643

PR #498, opened 2023-04-27
Localgw

jlin-nf:
Add guide for using ziti-router or ziti-tunnel as local gw


dovholuknf:
I'd really prefer to see us use text blocks and not screen caps for cli examples.

jlin-nf:
> I'd really prefer to see us use text blocks and not screen caps for cli examples.

converted cli image to text

r-caamano (approved):

dovholuknf (approved):
@r-caamano approved. approving so it can merge

dovholuknf:
I can't merge this PR. The last few commits were not signed. Unfortunately there's no 'easy' way to fix that. imo the easiest way will be to make a whole branch, merge without committing, and make a new PR with the commits signed.

PR #515, opened 2023-05-10
document a tun DOWN failure mode

### LINKED ISSUE #514: document Linux capabilities failure mode

qrkourier:
resolves #514

### DIFF
--- a/docusaurus/docs/reference/tunnelers/linux/linux-tunnel-troubleshooting.md
+++ b/docusaurus/docs/reference/tunnelers/linux/linux-tunnel-troubleshooting.md
@@ -50,3 +50,43 @@ If the tunneller is crashing then it may be crucial to collect and analyze the c
   In this case the dump is handled by `apport` which saves the file in `/var/crash`. I'll need to follow the `apport` documentation to learn how to unpack and parse the dump file.
 
 Please raise [a GitHub issue](https://github.com/openziti/ziti-tunnel-sdk-c/issues/) if you experience a crash.
+
+## tun device is DOWN
+
+If the tun device is DOWN then the tunneller will not be able to establish a connection to the Ziti network. You can check the status of the tun device with the `ip` command.
+
+```bash
+$ ip link sh tun0
+202: tun0: <POINTOPOINT,MULTICAST,NOARP> mtu 1500 qdisc noop state DOWN mode DEFAULT group default qlen 500
+    link/none 
+```
+
+Potential causes include override capabilities are set on the `/bin/ip` command.
+
+  ```bash
+  $ attr -l /bin/ip
+  Attribute "capability" has a 20 byte value for /bin/ip
+
+  $ getcap /bin/ip
+  /bin/ip =
+  ```
+
+An empty set of capabilities prevents inheriting ambient capabilities when invoked by systemd. You can remove the empty set with the `setcap -r` command.
+
+  ```bash
+  sudo setcap -r /bin/ip
+  ```
+
+Now the capabilities are completely removed.
+
+  ```bash
+  $ attr -l /bin/ip
+
+  $ getcap /bin/ip
+  ```
+
+Restart the tunneller service.
+
+  ```bash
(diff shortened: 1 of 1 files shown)


docusaurus/docs/reference/tunnelers/linux/linux-tunnel-troubleshooting.md
```diff
@@ -50,3 +50,43 @@ If the tunneller is crashing then it may be crucial to collect and analyze the c
   In this case the dump is handled by `apport` which saves the file in `/var/crash`. I'll need to follow the `apport` documentation to learn how to unpack and parse the dump file.
 
 Please raise [a GitHub issue](https://github.com/openziti/ziti-tunnel-sdk-c/issues/) if you experience a crash.
+
+## tun device is DOWN
+
+If the tun device is DOWN then the tunneller will not be able to establish a connection to the Ziti network. You can check the status of the tun device with the `ip` command.
+
+```bash
+$ ip link sh tun0
```
dovholuknf, line 59: why document this? shouldn't it just be "if ziti-edge-tunnel isn't running things won't work"? i personally don't love it when underlay concepts creep in like this. this feels to me like it should be "check if ziti-edge-tunnel is running"... 

now, the fact that ZET doesn't detect network change events, is a different ball of wax and also shouldn't really be documented. but filing a issue for that detection for all platforms -- not just linux -- makes sense to me

qrkourier: Updated to clarify that the ZET service is running in this failure mode.

scareything: The main issue here is that collectively we spent a lot of time trying to understand what was preventing `ip` commands from working when spawned by `ziti-edge-tunnel` and the systemd unit specified the required `CAP_NET_ADMIN` capability.

We found that an empty set of capabilities on /bin/ip is not the same as _no_ capabilities on /bin/ip. We wanted to capture this info somewhere, lest someone comes upon it by some chance.

scareything: Btw, the tun device shows as DOWN when the capabilities on /bin/ip are empty because ziti-edge-tunnel spawns /bin/ip to bring the tun interface "up". I suppose if we keep this gem of wisdom in the doc it would be worth mentioning the ERROR that is seen in the log when the ip command fails.

dovholuknf: Is this a really common problem people are seeing? I've never seen someone have this problem yet. If not, it feels like perhaps this could be boiled down into something else? Maybe this should be an issue filed on the ziti-edge-tunnel repo that we should file and 'fix' somehow?

qrkourier: Added the log message that is symptomatic of the tun DOWN condition

dovholuknf: Talked to Shawn a bit on this matter. To me, this doesn't seem to be the sort of thing that needs to be captured in the doc. If we're going to have a troubleshooting guide, it really should be more in the form of "if you see this, then then" type of thing. As it is, this is more of a "you have already determined the problem, here are some things to check". That doesn't feel like a troubleshooting guide to me.

I still think this should just be turned into an issue and handled in the tunneler.

qrkourier: @dovholuknf Your point is taken about if this, then that. The short version of my thoughts on that is that it makes the most sense to document the symptoms that are _most distinctive_ for a particular failure mode and describe the solution from that perspective.

qrkourier:
I will close this issue because it is too rare to document in the troubleshooting guide for the Linux tunnel.

PR #518, opened 2023-05-16
move digital ocean guide to more generic public cloud guide

jlin-nf:


r-caamano (commented):
Why is 2.1 Create a VM on Digital Ocean in the router section shouldn't this be generic with tabs for specific clouds like other items?

jlin-nf:
> Why is 2.1 Create a VM on Digital Ocean in the router section shouldn't this be generic with tabs for specific clouds like other items?

Fixed.

PR #530, opened 2023-06-01
clarify the steps to add an enrollment token

qrkourier:


qrkourier:
This is to reduce confusion about placing an enrollment token into the directory used by the Linux packages loading all identities from an `--identity-dir`.

docusaurus/docs/reference/tunnelers/linux/index.md
```diff
@@ -149,9 +149,11 @@ sudo apt install ziti-edge-tunnel
 
   ```bash
   sudo -u ziti tee /opt/openziti/etc/identities/ziti-id.jwt >/dev/null
-  # paste the contents of the enrollment token JWT file and press Ctrl+D
+  # run the above command then paste the token from clipboard then press Ctrl+D to close the file
```
dovholuknf: seems like we should just tell the user to put it somewhere of their choosing. also the whole ctrl+d thing, i get why you're doing it but i'd prefer to keep it all simpler... download the file... use it...

qrkourier: True, it doesn't _need_ to be in the standard path for a manual install.

qrkourier: I was confused and thought the confusion was about manual installation, but really this step to create the JWT is common to manual and package installs. The main difference between those is whether or not the standard path is important. I'll send another commit.

qrkourier:
forgot about this languishing PR and ended up doing it another way for https://github.com/openziti/ziti-doc/issues/579

PR #580, opened 2023-07-25
clarify placing the enroll token on Linux

### LINKED ISSUE #579: Use more familiar commands in the Linux install instructions
Feedback received is the one step example with `sudo -u ziti tee` is not intuitive and it would be better to use more familiar commands with plain purpose like `sudo -u ziti vi`, then paste, save, and exit. 

From forum topic: https://openziti.discourse.group/t/do-i-need-to-create-a-ziti-user-on-debian/1422/3?u=qrkourier

qrkourier:
resolves #579

### DIFF
--- a/docusaurus/docs/reference/tunnelers/linux/index.md
+++ b/docusaurus/docs/reference/tunnelers/linux/index.md
@@ -148,8 +148,8 @@ sudo apt install ziti-edge-tunnel
 2. Place an enrollment token JWT file or identity config JSON file in `/opt/openziti/etc/identities`.
 
   ```bash
-  sudo -u ziti tee /opt/openziti/etc/identities/ziti-id.jwt >/dev/null
-  # paste the contents of the enrollment token JWT file and press Ctrl+D
+  sudo -u ziti vi /opt/openziti/etc/identities/ziti-id.jwt
+  # paste the contents of the enrollment token JWT on a single line, save, and exit
   ```
 
 2. Enable and start the service
@@ -264,8 +264,8 @@ repo_gpgcheck=1
 2. Place an enrollment token JWT file or identity config JSON file in `/opt/openziti/etc/identities`.
 
   ```bash
-  sudo -u ziti tee /opt/openziti/etc/identities/ziti-id.jwt >/dev/null
-  # paste the contents of the enrollment token JWT file and press Ctrl+D
+  sudo -u ziti vi /opt/openziti/etc/identities/ziti-id.jwt
+  # paste the contents of the enrollment token JWT on a single line, save, and exit
   ```
 
 2. Enable and start the service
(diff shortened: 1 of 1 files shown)


docusaurus/docs/reference/tunnelers/linux/index.md
```diff
@@ -145,11 +145,12 @@ sudo apt install ziti-edge-tunnel
...
 
-2. Place an enrollment token JWT file or identity config JSON file in `/opt/openziti/etc/identities`.
+2. Place a copy of the enrollment token JWT file in the `/opt/openziti/etc/identities` directory.
+2. Set the filemode and owner so that members of group `ziti` will be able to read and delete the token file.
 
   ```bash
-  sudo -u ziti tee /opt/openziti/etc/identities/ziti-id.jwt >/dev/null
-  # paste the contents of the enrollment token JWT file and press Ctrl+D
+  sudo chown -cR :ziti        /opt/openziti/etc/identities
+  sudo chmod -cR ug=rwX,o-rwx /opt/openziti/etc/identities
```
dovholuknf, line 153: why doesn't the installer take care of this? feels like it should to me. Also is there any way to condense all these ubuntu's into a singular example? They all appear __nearly identical__ except for one word (the distro)? Seems redundant to me?

dovholuknf (approved):

docusaurus/docs/reference/tunnelers/linux/index.md
qrkourier: The installer sets the filemode on the dir. This instructs the user to create a new file, the enrollment JWT, which also needs to have permissions set for user "ziti." The easiest way to do that is set correct permissions recursively for the identities dir. Otherwise it would be necessary to either know the JWT filename in advance, or provide a placeholder like token.jwt which requires the user to substitute their own filename, an unnecessary step.

qrkourier: > any way to condense all these ubuntu's into a singular example

This would probably work for the Ubuntus.

```bash
(
set -euxo pipefail

curl -sSLf https://get.openziti.io/tun/package-repos.gpg \
  | sudo gpg --dearmor --output /usr/share/keyrings/openziti.gpg

source /etc/os-release

echo 'deb [signed-by=/usr/share/keyrings/openziti.gpg] https://packages.openziti.org/zitipax-openziti-deb-stable ${UBUNTU_CODENAME} main' \
  | sudo tee /etc/apt/sources.list.d/openziti.list >/dev/null

sudo apt update
sudo apt install ziti-edge-tunnel
)
```

qrkourier: proposed this consolidation in https://github.com/openziti/ziti-doc/pull/614

PR #596, opened 2023-07-27
NodeJS tidying - delete leftover files and document managing packages

qrkourier:


README.md
```diff
@@ -9,29 +9,31 @@
...
 
-```
-NOTE: you must be configured to run NodeJS >=16.14 before running the next command.
-```
+> [!NOTE]\
+> you must configure NodeJS >=16.14 before running `yarn`
 
-You can then run `cd ./docusaurus && yarn install && yarn start` to serve the Docusaurus site from webpack. If you're testing configuration changes you will need to serve the production build with `yarn serve` instead.
+> [!WARNING]\
+> It is never necessary to run `npm` to build this project or add a plugin. Always use `yarn`.
```
dovholuknf: > It is never necessary to run `npm`

change this to "Do not use `npm`

package-lock.json
dovholuknf, line 1: add these to the .gitignore too along with a comment in the .gitignore to "use yarn not npm"

docusaurus/package.json
```diff
@@ -21,6 +21,7 @@
     "@docusaurus/theme-mermaid": "^2.4.1",
     "@hotjar/browser": "^1.0.9",
     "@mdx-js/react": "^1.6.22",
+    "caniuse-lite": "^1.0.30001517",
```
dovholuknf: was this strictly necessary? was it used in node but not yarn's files? would you verify it's _necessary_ and if not remove it?

qrkourier: Fixed in branch. It was a mistake for me to declare the explicit dependency on caniuse-lite. I did that while I was figuring out how to upgrade the package with `yarn` and never went back to correct it.

.gitignore
```diff
@@ -17,3 +17,6 @@ docs-local
 
 github_deploy_key
 sessions
+
+# use yarn, not npm, see README.md for details
+package-lock.json
```
dovholuknf, line 22: and package.json too, right?

dovholuknf: and docusaurus/package-lock.json ?

qrkourier: This way Git will ignore files named `package-lock.json` no matter where they appear in the repo because the match pattern is not constrained by a path prefix. It's synonymous with `/**/package-lock.json`.

dovholuknf (approved):

Issue #151, opened 2022-08-17
Fully document how to 'bring your own pki'


dovholuknf:
cover big topics like where you can get a cert, process, what features each cert needs etc. make it a 'guide' or 'article' or something along those lines

qrkourier:
Couple of clues about this in articles with `alt_server_certs` https://openziti.github.io/search/?q=alt_server_certs but I take it you're thinking even bigger about bring-your-own-CA, not just server certs from a third-party CA.

dovholuknf:
This is a very big topic, so yes. it's probably much more than just that

qrkourier:
I've started working toward this goal in https://github.com/openziti/ziti-doc/pull/604 and https://github.com/openziti/ziti-doc/pull/592.

PR #604, opened 2023-07-31
update alt server certs guide

qrkourier:


qrkourier:
This article has a scope problem. It should be renamed like "Self-Hosting the BrowZer Bootstrapper," and stop generalizing "public certificates." The current scope is not truly a "guide" because it tries to field FAQs like "do I need 3rd-party server certs on my edge APIs?" that would be moot if the guide focused on the object of configuring BrowZer.

qrkourier:
@dovholuknf ready for review

dovholuknf:
Really? Looks like you're just totally removing the file not "updating" it?

qrkourier:
This rewrite was a failed attempt to partially answer [the issue about documenting bring-your-own-pki](https://github.com/openziti/ziti-doc/issues/151). We should use that issue to organize and focus efforts on the various topics the issue represents. It's a broad issue, so we should peel off pieces of that which can be addressed with a particular type of doc, like a guide, explainer, or tutorial.

dovholuknf:
~~I don't know what the thumbs up emoji means. Is the _plan_ to remove this file entirely?~~ Your comment and mine came in at the same time...

dovholuknf (approved):

PR #612, opened 2023-08-02
start a style guide, small tweaks to readme, try some frontmatter

dovholuknf:


qrkourier:
My wishlist for the style guide includes the lessons we've learned about using Docusaurus, like using file paths for links.  I would love to see some Markdown guidance too, even something general like "be consistent with how you use Markdown." We could also provide specific guidance like "run `markdownlint --fix ./my/changed/file.md` to conform new/changed files to this project's `.markdownlint.yaml`.

PR #619, opened 2023-08-03
add guide for wildcard dns

qrkourier:


docusaurus/docs/guides/wildcard-dns-kerberos.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Wildcard DNS and Kerberos
+sidebar_label: Wildcard DNS
+sidebar_position: 50
+---
+
+### How to Tunnel Kerberos/Active Directory DNS Queries
+
+Kerberos/AD domain members can communicate with their domain controller (DC) using a Ziti network. This involves forwarding DNS queries to the DC's nameservers with a Ziti service.
```
dovholuknf, line 9: This doesn't feel very 'guide-like' to me. it's providing me a bunch of ancillary information and not "here is how you do it" which is what I would expect to see from a guide. Do this do that etc. It's not feeling like 'a guide' to me

docusaurus/docs/guides/wildcard-dns-kerberos.md
```diff
@@ -0,0 +1,19 @@
...
+
+The Ziti tunneler's built-in nameserver behaves differently when it is listening for DNS queries that match a wildcard pattern. When it's listening for an exact match, only queries for record type A are considered. When it's listening for a wildcard pattern, e.g., `*.my.kerberos.domain.example.com`, queries of type A, MX, TXT, and SRV are considered. Only the least significant label of the query name may be a wildcard.
+
+Any match for an A record is answered with an intercept IPv4 address that routes to the Ziti tunneler's network interface. Matching queries for record types MX, TXT, and SRV are proxied to the hosting tunneler. The hosting tunneler will then resolve the query according to its host's default resolver configuration. The response is returned to the intercepting tunneler which then responds to the original query.
```
scareything, line 13: ~~intercept~~

docusaurus/docs/guides/wildcard-dns-kerberos.md
```diff
@@ -0,0 +1,19 @@
...
+title: Wildcard DNS and Kerberos
+sidebar_label: Wildcard DNS
+sidebar_position: 50
+---
+
+### How to Tunnel Kerberos/Active Directory DNS Queries
+
+Kerberos/AD domain members can communicate with their domain controller (DC) using a Ziti network. This involves forwarding DNS queries to the DC's nameservers with a Ziti service.
+
+The Ziti tunneler's built-in nameserver behaves differently when it is listening for DNS queries that match a wildcard pattern. When it's listening for an exact match, only queries for record type A are considered. When it's listening for a wildcard pattern, e.g., `*.my.kerberos.domain.example.com`, queries of type A, MX, TXT, and SRV are considered. Only the least significant label of the query name may be a wildcard.
```
scareything, line 11: Maybe this needs some wordsmithing? I'm not sure what this sentence is trying to get across but the dns server is always "listening" for exact matches and for wildcard patterns; in fact the dns server doesn't really know the difference. The query type doesn't matter either. If we look up a hostname that happens to be in a domain that's been designated as a "wildcard" domain (via an intercept config with an address that starts with "*."), then we will manufacture an IP address to return for this hostname, and store the result internally so the same answer is given the next time the name is queried.

PR #630, opened 2023-08-07
Add style guide local (and some link unfurling/seo)

dovholuknf:
some link unfurling and seo has crept into this pr as well


STYLEGUIDE.md
```diff
@@ -0,0 +1,61 @@
...
+
+Here is a list of expectations you should follow when writing a page backed by Docusaurus. These are in no particular
+order at this time but should be at least roughly grouped :
+
+* markdown: __DO__ wrap lines or format lines ___you are adding___ Consider wrapping at around 120 chars
+* markdown: __DO NOT__ wrap lines for content ___you are not touching___
+* markdown: __DO NOT__ format or wrap lines for an entire file automatically
+* markdown: __DO__ add [front matter](https://docusaurus.io/docs/next/markdown-features#front-matter) to every markdown
+            page. __Most importantly__, `title` and `description` but optionally consider adding and `image`
+* format: __DO__ format an entire markdown and NOTHING else (if trying to format a document)
```
qrkourier: I think this means "If formatting a Markdown document, do send the edits in a separate pull request."

STYLEGUIDE.md
```diff
@@ -0,0 +1,61 @@
...
+* format: __DO NOT__ format an entire markdown __and change__ content, format content __ONLY__ and with agreement from others
```
qrkourier, line 28: Should this be two separate guidelines? It seems to be saying both "Send separate pull requests for formatting changes" and "Don't send pull requests for content changes, but instead discuss them first in Discourse."

STYLEGUIDE.md
```diff
@@ -0,0 +1,61 @@
...
+* links: __DO__ use relative path links to the actual markdown file (../../../docs/some/05-markdown.md)
+* links: __DO NOT__ use fully qualified links pointing to the docs (https://openziti.io/...)
+* links: __DO NOT__ use the Docusaurus 'id' in a link as in: `../../../docs/some/markdown`. (Notice the lack of ".md"/".mdx")
+* images: __DO__ put images into the `/docusaurus/static/img` directory
```
qrkourier, line 32: Is this better than placing non-Markdown assets in the same folder as the Markdown that references them?

STYLEGUIDE.md
```diff
@@ -0,0 +1,61 @@
...
+* images: __DO__ try to organize and name images properly
+* images: __DO__ try to use `.SVG` files as much as possible
```
qrkourier, line 34:
```suggestion
* images: __DO__ try to use `.SVG` files as much as possible
* diagrams: __DO__ use ` ```mermaid ` in-line fencing instead of images or links to `mermaid.ink`
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,61 @@
...
+* images: __DO__ take care to test your image on both light and dark mode to see if it looks acceptable in both views
+* images: __DO__ try to not push too many binary (image) files over and over as it bloats the git repo
+* sidebar: __DO__ try to favor organizing pages using numerical prefixing: `01-page1.md`, `10-page2.md`
```
qrkourier, line 37: This is supported by Docusaurus, but inferior to sidebar position metadata in practice because it creates work when the index prefix must change thereby breaking file path references in Markdown links, and because the index prefix is subordinate to position metadata, leading to a non-obvious precedence of prefixes and metadata competing for the end result.

STYLEGUIDE.md
```diff
@@ -0,0 +1,61 @@
...
+* sidebar: __DO__ add distance between numerical prefixing to allow for easier insertion of pages (since this is a
+           pain to fix). This is done _specifically_ to allow the file system to reflect the view on site.
+* sidebar: __DO__ add a landing page to any sidebar groups via `_category_.json`. Example:
```
qrkourier, line 40: This isn't always best. For example, an overview page may be desired for the landing page, but adding it as the category index will hide it from the sidebar. Alternatives include the generated index cards page.

STYLEGUIDE.md
```diff
@@ -0,0 +1,61 @@
...
+           `"link": { "type": "doc", "id": "reference/tunnelers/index" }`
+
+## Types of Doc
```
qrkourier, line 44: Yay!

docusaurus/docs/learn/introduction/index.mdx
```diff
@@ -1,6 +1,8 @@
 ---
 title: What is OpenZiti?
 hide_table_of_contents: true
+description: OpenZiti is an open source zero trust network overlay built on the notion that everything is an SDK
```
qrkourier, line 4:
```suggestion
description: OpenZiti is an open-source zero trust network overlay built on the notion that everything is an SDK
```

qrkourier (dismissed):

qrkourier:
We could also provide specific guidance like "run `markdownlint --fix ./my/changed/file.md` to conform new/changed files to this project's `.markdownlint.yaml`.

dovholuknf:
> We could also provide specific guidance like "run `markdownlint --fix ./my/changed/file.md` to conform new/changed files to this project's `.markdownlint.yaml`.

I'm not looking to add that sort of thing to this guide at this time

STYLEGUIDE.md
dovholuknf: it's saying "don't just willy nilly format content, only format content you are not touching with agreement from all". aka, don't just change formatting because your editor/process wants to

STYLEGUIDE.md
dovholuknf: yes. that's why i added it? do not put binary content into the same folder as the markdown

STYLEGUIDE.md
qrkourier: Re: agreement

Is sending a pull request a good way to propose a change and ask for that agreement?

STYLEGUIDE.md
dovholuknf: yep. it's also more clear and easier to see how it sorts. let's hope we don't find ourselves reorganizing things too much.

STYLEGUIDE.md
dovholuknf: No. It's far too late. Any/all prs run the risk of having conflicts at that point. formatting markdown, like code, should only be done 'as you go' (which is relatively easy/risk free) or when there's nothing else going on to avoid any conflicts

STYLEGUIDE.md
dovholuknf: The generated index cards are generally useless. It's better to have a bespoke page per group, describing what the group is for etc.

STYLEGUIDE.md
dovholuknf: i'll make this commit.

STYLEGUIDE.md
qrkourier: My primary objection is that it breaks links when prefixes change. If we don't need to reorganize things too much that's less of an issue. My secondary objection is that it is less clear, and harder to see how it sorts because the prefixes are subordinate to position metadata. The only way to make prefixes clear is to prohibit the use of position metadata, which is not suggested by these guidelines.

dovholuknf: it's superior to see how it sorts, it's sorted in the directory. that's the major benefit. the major drawback is that it's a major pain to rearrange things dramatically. that's why, just like the BASIC days, if you leave sensible gaps, rearranging things is not as big of an issue (one hopes)

I would definitely prefer to prohibit position data, but instead of being overly hamfisted, I figure sensibilities are allowable. using front matter, while flexible, is incredibly difficult to debug. The other option is to explicitly manage the sidebar. If you would prefer to mandate the prohibition of frontmatter from positioning, that's fine with me. i can add that too. i just didn't feel the need for it

qrkourier: I have wondered if the autogenerated sidebar is more or less trouble than a manual sidebar. I'm comfortable with leaving both indexing and position metadata as tools in the toolbox. I recommend that we do not recommend indexing. That's overly prescriptive and will create unnecessary churn. Indexing has it's place but I'm not in favor of prescribing that moving forward nor retrofitting  the site with indexed content filenames, heaven forbid.

dovholuknf: > I'm comfortable with leaving both indexing and position metadata as tools in the toolbox.

I am not. It leads to confusion. Until it creates unnecessary churn, I am going to add it to the guide. People aren't mandated to follow it anyway so we'll see what happens. We will not be refactoring anything that's already out there.

STYLEGUIDE.md
```diff
@@ -0,0 +1,62 @@
...
+order at this time but should be at least roughly grouped :
+
+* markdown: __DO__ wrap lines or format lines ___you are adding___ Consider wrapping at around 120 chars
+* markdown: __DO NOT__ wrap lines for content ___you are not touching___
+* markdown: __DO NOT__ format or wrap lines for an entire file automatically
+* markdown: __DO__ add [front matter](https://docusaurus.io/docs/next/markdown-features#front-matter) to every markdown
+            page. __Most importantly__, `title` and `description` but optionally consider adding and `image`
+* format: __DO__ format an entire markdown and NOTHING else (do not reformat and change content!)
+* format: __DO NOT__ format an entire markdown __and change__ content, format content __ONLY__ and with agreement from others
+* links: __DO__ use relative path links to the actual markdown file (../../../docs/some/05-markdown.md)
```
qrkourier, line 29:
```suggestion
* links: __DO__ use absolute path links to the actual markdown file (`/learn/some/05-markdown.md`). This absolute link corresponds to the Markdown file in `/docusaurus/docs/learn/some/05-markdown.md`. Note that the absolute link starts with the first directory name under the Docusaurus content plugin directory, e.g. `docs/`.

```

STYLEGUIDE.md
```diff
@@ -0,0 +1,62 @@
...
+* links: __DO NOT__ use fully qualified links pointing to the docs (https://openziti.io/...)
+* links: __DO NOT__ use the Docusaurus 'id' in a link as in: `../../../docs/some/markdown`. (Notice the lack of ".md"/".mdx")
+* images: __DO__ put images into the `/docusaurus/static/img` directory
+* images: __DO__ try to organize and name images properly
```
qrkourier, line 33: Define properly?

STYLEGUIDE.md
```diff
@@ -0,0 +1,62 @@
...
+* images: __DO__ try to use `.SVG` files as much as possible
```
qrkourier, line 34:
```suggestion
* images: __DO__ use SVG images instead of PNG or JPEG whenever possible.

```

qrkourier (changes requested):

PR #631, opened 2023-08-07
compare main..HEAD for Vercel ignore command

qrkourier:


dovholuknf (dismissed):

qrkourier:
The clone is shallow, and should be 10 commits deep, yet Vercel is complaining that only commit index 0 is available. It's not clear why. They have new features for ignoring builds that might work better than the shell one-liner we've been using.

dovholuknf (approved):

PR #658, opened 2023-08-24
First pass at host.v2 config reference doc

plorenz:


dovholuknf:
there's host.v1 and intercept.v1 at /docs/learn/core-concepts/config-store/overview. seems like those could be moved down here?

docusaurus/docs/reference/config-types/host.v2.md
```diff
@@ -0,0 +1,291 @@
...
+---
+
+# The `Host.v2` Config Type
+
+The `host.v2` configuration type defines how edge routers or tunnelers should create terminators and
+health checks for the application endpoints that provide a service.
+
+## Terminators
+
+Each config can define one or more terminators. If a service is hosted by multiple servers, the
```
dovholuknf: I personally think "terminators" is a potentially confusing term for the reader. I know that's the field name, so referencing it as such makes sense to me. The "terminators" array is really an array of `host.v1` entries, right? I'd advocate that we simplify the host.v2 by putting all this nice detail into the `host.v1` structure/doc and making this page be more of a "this structure exists to create/aggregate host.v1 entities".

docusaurus/docs/reference/config-types/host.v2.md
```diff
@@ -0,0 +1,291 @@
...
+    }
+  ]
+}
+```
+
+Services that are forwarding traffic from an OpenZiti tunneling proxy may use the following
+properties to indicate what should be forwarded:
+
+* `forwardProtocol` - flag indicating that the protocol of the forwarded connection is to be used
+    * Must be the value `true`
```
dovholuknf: > Must be the value `true`

reads awkwardly. Maybe "Can only be set to true" or "If set, the value must only be set to true" and then probably not in bulletted form?

docusaurus/docs/reference/config-types/host.v2.md
```diff
@@ -0,0 +1,291 @@
...
+      "allowedProtocols": [
+        "tcp",
+        "udp"
+      ]
+    }
+  ]
+}
+```
+
+Note that not everything must be forwarded. For example the address could be hard coded, while the
```
dovholuknf: > not everything must be forwarded

Maybe, something like: "Note that the address is not forwarded in the example below. The port and protocol are forwarded and the corresponding 'allow' is set, but the address is statically set to '192.168.100.1" ?

docusaurus/docs/reference/config-types/host.v2.md
```diff
@@ -0,0 +1,291 @@
...
+      ]
+    }
+  ]
+}
+```
+
+Health checks and listen options also can be specified for each terminator.
+
+* `listenOptions` - Provides ways to customize the terminator
+    * See below for the full definition
```
dovholuknf: You can use anchor links here: `#listen-options` etc and avoid the "see below" bullet i think

docusaurus/docs/reference/config-types/host.v2.md
```diff
@@ -0,0 +1,291 @@
...
+## Health Checks
+
+There are two kinds of health checks supported, port check and http checks.
+
+### Port Checks
+
+Port checks verify if a given port is accepting TCP connections. The check does not attempt to send
+or receive any data. Port check definitions support the following properties:
+
+* `address` - an IP or DNS address with port.
```
dovholuknf: you capitalized IP here but didn't a couple of times above.  Probably best to use IP and IPs?

docusaurus/docs/reference/config-types/host.v2.md
```diff
@@ -0,0 +1,291 @@
...
+    * This field is required.
```
dovholuknf: maybe change to:
```
* `address` __[REQUIRED]__ - an IP or...
```
just to shrink it up some?

docusaurus/docs/reference/config-types/host.v2.md
```diff
@@ -0,0 +1,291 @@
...
+    * Valid actions include:
+        * `mark unhealthy` - sets the associated terminator's precedence to `failed`.
+        * `mark healthy` - sets the associated terminator's precedence back from `failed` to its
+          original value.
+        * `increase cost N` - increases the cost of the associated terminator by `N`.
+        * `decrease cost N` - decreases the cost of the associated terminator by `N`.
+        * `send event` - causes a terminator event to be emitted from the controller. Useful for
+          alerting or external integrations.
+
+:::note Although multiple health checks can be configured, it's best if the actions don't overlap.
```
dovholuknf: put the `:::note` on its own line else it renders awkwardly. you can use this to change the title though instead of "Info" or "note" etc
image

docusaurus/docs/reference/config-types/host.v2.md
plorenz: I'm tempted to allow aliasing `terminators` as `hosts`. Unfortunately, that would need to be either translate to terminators on save, which could be confusing, or would need to be handled in various places. Not sure if there's a good solution there.

Unfortunately, the host.v2 contents are almost, but not exactly, the same as host.v1. there are few places where some cleanup was done, so they're not 1:1.

docusaurus/docs/reference/config-types/host.v2.md
plorenz: We should make a standard for all ziti doc, since this is something that we need to notate in multiple places, and we should do it consistently

docusaurus/docs/reference/config-types/host.v2.md
plorenz: fixed

docusaurus/docs/reference/config-types/host.v2.md
plorenz: fixed

docusaurus/docs/reference/config-types/host.v2.md
plorenz: fixed

docusaurus/docs/reference/config-types/host.v2.md
plorenz: fixed

docusaurus/docs/reference/config-types/host.v2.md
plorenz: fixed

docusaurus/docs/reference/config-types/host.v1.md
```diff
@@ -0,0 +1,263 @@
+---
+sidebar_label: host.v1
+sidebar_position: 15
+---
+
+# The `host.v1` Config Type
+
+The `host.v1` configuration type defines how edge routers or tunnelers can make outgoing connections
+for associated services. The config type also allows specifying health checks for the service. An
+endpoint in a `host. v1` config can be specified explicitly or can be configured to use information
```
scareything: stray space in `host. v1` and `host. v2` appears often (but not always) in this doc.

docusaurus/docs/reference/config-types/host.v1.md
```diff
@@ -0,0 +1,263 @@
...
+  "protocol": "tcp"
+}
+```
+
+## Forwarding Configuration
+
+Services that are forwarding traffic from an OpenZiti tunneling proxy may use the following
+properties to indicate what should be forwarded:
+
+* `forwardProtocol` - flag indicating that the protocol of the forwarded connection is to be used.
```
scareything: imo instead of referring to the _forwarded_ connection, I think it would be more correct to describe the `forward*` fields as "indicating that the (protocol|port|address) of the _intercepted_ connection is to be used.

or perhaps more succinctly: connect to the (protocol|port|address) that the intercepted client was connecting to

docusaurus/docs/reference/config-types/host.v1.md
```diff
@@ -0,0 +1,263 @@
...
+  ],
+  "forwardProtocol": true,
+  "allowedProtocols": [
+    "tcp",
+    "udp"
+  ]
+}
+```
+
+Health checks and listen options also can be specified for each terminator.
```
scareything: this is the first use of _terminator_. Maybe an association between host.vX and terminator is made elsewhere, but I didn't see it. I think it would be better to just mention `host.vX` here. e.g. "... also can be specified in host.v1/host.v2 configurations".

docusaurus/docs/reference/config-types/host.v2.md
```diff
@@ -0,0 +1,50 @@
...
+---
+
+# The `host.v2` Config Type
+
+The `host.v2` configuration type defines how edge routers or tunnelers can make outgoing connections
+for associated services. The config type also allows specifying health checks for the service. An
+endpoint in a `host. v2` config can be specified explicitly or can be configured to use information
+passed through from a tunneler intercepting traffic, allowing it to forward traffic.
+
+The `host. v2` type allows defining multiple host endpoints. If only a single endpoints is required,
```
scareything: single _endpoint_

docusaurus/docs/reference/config-types/host.v1.md
```diff
@@ -0,0 +1,263 @@
...
+  Can only be set to true.
+* `allowedProtocols` - the list of allowed protocols. Valid values include `tcp` and `udp`
+* `forwardAddress` - flag indicating that the target address of the forwarded connection is to be
+  used. Can only be set to true.
+* `allowedAddresses` - the list of allowed addresses. Valid values include IPs, hostnames and CIDRs
+* `forwardPort` - flag indicating that the target port of the forwarded connection is to be used.
+  Can only be set to true.
+* `allowedPortRanges` - the list of allowed port ranges.
+    * Example: `allowedPortRanges: [ {"low" : 80, "high" : 80}, {"low" : 8080, "high" : 8090} ]`
+* `allowedSourceAddresses` - list of addresses in IP, hostname or CIDR
```
scareything: list of _allowed_ addresses... ?

scareything (dismissed):

docusaurus/docs/reference/config-types/host.v1.md
plorenz: fixed

docusaurus/docs/reference/config-types/host.v1.md
plorenz: fixed

docusaurus/docs/reference/config-types/host.v1.md
plorenz: I removed references to `terminator` for now. Once we have a entity reference section, I think we can link to that and explain how they tie together.

docusaurus/docs/reference/config-types/host.v2.md
plorenz: fixed

docusaurus/docs/reference/config-types/host.v1.md
plorenz: fixed

scareything (approved):

PR #660, opened 2023-08-26
use absolute Markdown links 

qrkourier:
...and add a script to convert relative links to absolute


check-links/markdown-absolute-file-links.bash
```diff
@@ -0,0 +1,30 @@
...
+    exit 1
+fi
+
+# filter for markdown files that contain relative links to other markdown files
+findMarkdownFiles(){
+    find "$1" \
+        -regextype posix-extended \
+        -regex ".*\.mdx?" \
+        -print0 \
+        | xargs -0r grep -lE '\]\(\.\./[^)]+\.mdx?\)'
```
qrkourier: Flaw: this fails to find Md files that only contain links with anchor fragments

qrkourier: It's probably best to stop filtering the found Md files because and it's redundant to the regex that extracts relative links.

gberl002 (dismissed):

check-links/markdown-absolute-file-links.bash
qrkourier: @gberl002 I found a few more after fixing a bug in the script that caused it to ignore Markdown files that contained only fragment/anchor links.

gberl002 (approved):

PR #664, opened 2023-08-31
Adds option to manage miniziti hosts file entries. Adds missing log level

sabedevops:


docusaurus/docs/learn/quickstarts/network/miniziti.bash
```diff
@@ -136,6 +182,14 @@ logInfo() {
     echo "INFO: $*"
 }
 
+logWarn() {
+    (( $# )) || {
+        logError "logWarn() takes 1 or more args"
+        return 1
+    }
+    echo "WARN: $*"
```
qrkourier:
```suggestion
    echo "WARN: $*" >&2
```

qrkourier (changes requested):
WARN messages to stderr, and I don't immediately love the pattern of kill switch vars in the called functions. My reflex is to make that less surprising. Do you feel strongly it should be that way?

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: Fixed. This function was missing but was being called during delete, so it wasn't an attempt at an original implementation, but a copy-paste of the info variant.

sabedevops: What is a "kill switch var"? Are you referring to lines 141 and 156?

qrkourier: I meant the global vars in the functions that are tested when the function is run. 

If they evaluate true then the function runs, else the function returns an error. 

Normally, I was surprised to find a "kill switch" like that inside the function. I expected it to check whether the function should run before calling it.

qrkourier:
I don't remember if there was a particular reason this approach couldn't work on macOS, and it's unclear whether there's demand for this on macOS. I'd say keep it Linux-only until there's demand.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: Updated to make the check in the caller. FWIW, calling something from a `main()` function and having it check the global there is not really any different. I used `if` instead of `&&` here due to issues with syntax highlighters in nested statements with the following shape:

```bash
(( test1 )) && {
  (( test2 )) && statement
}
```
Originally, this was the movitivation for moving them into the function in the first place.

sabedevops:
> I don't remember if there was a particular reason this approach couldn't work on macOS, and it's unclear whether there's demand for this on macOS. I'd say keep it Linux-only until there's demand.

I don't have an immediate means or motivation to  set up testing on macOS. Happy to add it as an option if you want it.

qrkourier (approved):

PR #671, opened 2023-09-06
Feature/multi profile support

sabedevops:
This PR attempts to deliver on the Miniziti promise of multiple Ziti profile support. To support that, this PR additionally adds some useful (context-aware) utility commands helpful in a multi-profile paradigm.

Features:
* `login` command
* `show-admin` command
* `ziti` command wrapper
* XDG-based per-profile namespacing
* Richer logging
* Idempotent management of `/etc/hosts` entries

### DIFF
--- a/docusaurus/docs/learn/quickstarts/network/miniziti.bash
+++ b/docusaurus/docs/learn/quickstarts/network/miniziti.bash
@@ -97,7 +97,7 @@ detectOs(){
 
 makeMinizitiStateDir() {
     case "$DETECTED_OS" in
-        MacOS)
+        macOS)
             state_dir="${XDG_STATE_DIR:-$HOME/Library/Application Support}/miniziti"
             ;;
         Linux|Windows)
@@ -224,16 +224,17 @@ logger() {
     fi
 
     caller_level="${caller##log}"
+    level="$(echo "$caller_level" | tr '[:lower:]' '[:upper:]')"
     if (( MINIZITI_DEBUG )); then
-        line="${caller_level^^} ${FUNCNAME[2]}:${BASH_LINENO[1]}: $message"
+        line="$level ${FUNCNAME[2]}:${BASH_LINENO[1]}: $message"
     else
-        line="${caller_level^^} $message"
+        line="$level $message"
     fi
 
     if [[ -n "${raw_message-}" ]]; then
-        echo -E "${line}"
+        echo -E "$line"
     else
-        echo -e "${line}"
+        echo -e "$line"
     fi
 }
 
--- a/docusaurus/docs/learn/quickstarts/network/miniziti.bash
+++ b/docusaurus/docs/learn/quickstarts/network/miniziti.bash
@@ -29,8 +29,8 @@ _usage(){
     echo -e "\n COMMANDS\n"\
             "   start\t\tstart miniziti (default)\n"\
             "   delete\t\tdelete miniziti\n"\
+            "   creds\t\tprints admin user updb credentials\n"\
             "   login\t\trun ziti edge login with miniziti context\n"\
-            "   admin-creds\t\tprints admin user updb credentials\n"\
             "   ziti\t\tziti cli wrapper with miniziti context\n"\
             "   help\t\tshow these usage hints\n"\
             "\n OPTIONS\n"\
@@ -293,10 +293,10 @@ main(){
             delete)         DELETE_MINIZITI=1
                             shift
             ;;
-            login)          DO_ZITI_LOGIN=1
+            creds)          SHOW_ADMIN_CREDS=1
                             shift
             ;;
-            admin-creds)     SHOW_ADMIN_CREDS=1
+            login)          DO_ZITI_LOGIN=1
                             shift
             ;;
             ziti)           RUN_ZITI_CLI=1
--- a/docusaurus/docs/learn/quickstarts/network/miniziti.bash
+++ b/docusaurus/docs/learn/quickstarts/network/miniziti.bash
@@ -15,7 +15,7 @@ banner(){
 
 BANNER
 
-    if [[ -n "$profile" ]] && [[ "$profile" != "miniziti" ]]; then
+    if [[ -n "$profile" && "$profile" != "miniziti" ]]; then
         echo -e "   .: $* :.\n"
     fi
 }
@@ -365,9 +365,9 @@ main(){
 
     MINIZITI_PROFILE_MARKER="miniziti-controller.$MINIKUBE_PROFILE."
     MINIZITI_INGRESS_ZONE="$MINIKUBE_PROFILE.ziti"
-    state_dir="$(makeMinizitiXDGStateDir "$DETECTED_OS")"
-    profile_dir="$state_dir/profiles/${MINIKUBE_PROFILE}"
-    identities_dir="$profile_dir/identities"
+    STATE_DIR="$(makeMinizitiXDGStateDir "$DETECTED_OS")"
+    PROFILE_DIR="$STATE_DIR/profiles/${MINIKUBE_PROFILE}"
+    IDENTITIES_DIR="$PROFILE_DIR/identities"
 
     if (( DO_ZITI_LOGIN )); then
         miniziti_login
@@ -394,15 +394,15 @@ main(){
             cleanHosts
         fi
 
-        if [[ -d "$profile_dir" ]]; then
-            logWarn "Deleting miniziti profile directory: $profile_dir"
-            rm -rf  "$profile_dir"
+        if [[ -d "$PROFILE_DIR" ]]; then
+            logWarn "Deleting miniziti profile directory: $PROFILE_DIR"
+            rm -rf  "$PROFILE_DIR"
         fi
 
-        cert_file="$(find "$HOME/.config/ziti/certs" -maxdepth 1 -type f -name "miniziti-controller.$MINIKUBE_PROFILE.*" -print -quit)"
-        if [[ -n "$cert_file" ]]; then
-            logWarn "Deleting miniziti certificate file: $cert_file"
-            rm -f  "$cert_file"
+        CERT_FILE="$(find "$HOME/.config/ziti/certs" -maxdepth 1 -type f -name "miniziti-controller.$MINIKUBE_PROFILE.*" -print -quit)"
+        if [[ -n "$CERT_FILE" ]]; then
+            logWarn "Deleting miniziti certificate file: $CERT_FILE"
+            rm -f  "$CERT_FILE"
         fi
 
         logWarn "Removing $MINIKUBE_PROFILE profile identity from ziti-cli.json"
@@ -411,9 +411,9 @@ main(){
         exit 0
     }
 
-    if [[ ! -d "$identities_dir" ]]; then
-        logDebug "Creating miniziti identities directory: ($identities_dir)"
-        mkdir -p "$identities_dir"
+    if [[ ! -d "$IDENTITIES_DIR" ]]; then
+        logDebug "Creating miniziti identities directory: ($IDENTITIES_DIR)"
+        mkdir -p "$IDENTITIES_DIR"
     fi
 
     #
@@ -666,19 +666,19 @@ main(){
     logInfo "Setting default ziti identity to: $MINIKUBE_PROFILE"
     ziti_wrapper edge use "$MINIKUBE_PROFILE" >&3
 
-    router_name='miniziti-router'
-    router_ott="$identities_dir/$router_name.jwt"
-    if  ziti_wrapper edge list edge-routers "name=\"$router_name\"" \
+    ROUTER_NAME='miniziti-router'
+    ROUTER_OTT="$IDENTITIES_DIR/$ROUTER_NAME.jwt"
+    if  ziti_wrapper edge list edge-routers "name=\"$ROUTER_NAME\"" \
         | grep -q miniziti-router; then
-        logDebug "updating $router_name"
-        ziti_wrapper edge update edge-router "$router_name" \
+        logDebug "updating $ROUTER_NAME"
+        ziti_wrapper edge update edge-router "$ROUTER_NAME" \
             --role-attributes "public-routers" >&3
     else
-        logDebug "creating $router_name"
-        ziti_wrapper edge create edge-router "$router_name" \
+        logDebug "creating $ROUTER_NAME"
+        ziti_wrapper edge create edge-router "$ROUTER_NAME" \
             --role-attributes "public-routers" \
             --tunneler-enabled \
-            --jwt-output-file "$router_ott" >&3
+            --jwt-output-file "$ROUTER_OTT" >&3
     fi
 
     if  helm status ziti-router --namespace "${ZITI_NAMESPACE}" &>/dev/null; then
@@ -697,7 +697,7 @@ main(){
         }
         helm install "ziti-router" "${ZITI_CHARTS_REF}/ziti-router" \
             --namespace "${ZITI_NAMESPACE}" \
-            --set-file enrollmentJwt="$router_ott" \
+            --set-file enrollmentJwt="$ROUTER_OTT" \
             --set edge.advertisedHost="miniziti-router.${MINIZITI_INGRESS_ZONE}" \
             --set linkListeners.transport.advertisedHost="miniziti-router-transport.${MINIZITI_INGRESS_ZONE}" \
             --set "ctrl.endpoint=ziti-controller-ctrl.${ZITI_NAMESPACE}.svc:443" \
@@ -710,7 +710,7 @@ main(){
         --for condition=Available=True >&3
 
     logDebug "probing miniziti-router for online status"
-    if ziti_wrapper edge list edge-routers "name=\"$router_name\"" \
+    if ziti_wrapper edge list edge-routers "name=\"$ROUTER_NAME\"" \
         | awk '/miniziti-router/ {print $6}' \
         | grep -q true; then
         logInfo "miniziti-router is online"
@@ -757,29 +757,29 @@ main(){
     ## Ensure OpenZiti Identities and Services are Created
     #
 
-    client_name="${MINIKUBE_PROFILE}-client"
-    client_ott="$identities_dir/$client_name.jwt"
+    CLIENT_NAME="${MINIKUBE_PROFILE}-client"
+    CLIENT_OTT="$IDENTITIES_DIR/$CLIENT_NAME.jwt"
 
-    if  ! ziti_wrapper edge list identities "name=\"$client_name\"" --csv \
-        | grep -q "$client_name"; then
-        logDebug "creating identity $client_name"
-        ziti_wrapper edge create identity device "$client_name" \
-            --jwt-output-file "$client_ott" --role-attributes httpbin-clients >&3
+    if  ! ziti_wrapper edge list identities "name=\"$CLIENT_NAME\"" --csv \
+        | grep -q "$CLIENT_NAME"; then
+        logDebug "creating identity $CLIENT_NAME"
+        ziti_wrapper edge create identity device "$CLIENT_NAME" \
+            --jwt-output-file "$CLIENT_OTT" --role-attributes httpbin-clients >&3
     else
-        logDebug "ignoring identity $client_name"
+        logDebug "ignoring identity $CLIENT_NAME"
     fi
 
-    httpbin_name="httpbin-host"
-    httpbin_ott="$identities_dir/$httpbin_name.jwt"
-    httpbin_json="$identities_dir/$httpbin_name.json"
+    HTTPBIN_NAME="httpbin-host"
+    HTTPBIN_OTT="$IDENTITIES_DIR/$HTTPBIN_NAME.jwt"
+    HTTPBIN_JSON="$IDENTITIES_DIR/$HTTPBIN_NAME.json"
 
-    if  ! ziti_wrapper edge list identities "name=\"$httpbin_name\"" --csv \
-        | grep -q "$httpbin_name"; then
-        logDebug "creating identity $httpbin_name"
-        ziti_wrapper edge create identity device "$httpbin_name" \
-            --jwt-output-file "$httpbin_ott" --role-attributes httpbin-hosts >&3
+    if  ! ziti_wrapper edge list identities "name=\"$HTTPBIN_NAME\"" --csv \
+        | grep -q "$HTTPBIN_NAME"; then
+        logDebug "creating identity $HTTPBIN_NAME"
+        ziti_wrapper edge create identity device "$HTTPBIN_NAME" \
+            --jwt-output-file "$HTTPBIN_OTT" --role-attributes httpbin-hosts >&3
     else
-        logDebug "ignoring identity $httpbin_name"
+        logDebug "ignoring identity $HTTPBIN_NAME"
     fi
 
     if  ! ziti_wrapper edge list configs 'name="httpbin-intercept-config"' --csv \
@@ -845,17 +845,17 @@ main(){
         logDebug "ignoring service-edge-router-policy public-routers"
     fi
 
-    if [[ -s "$httpbin_ott" ]]; then
-        logDebug "enrolling $httpbin_ott"
+    if [[ -s "$HTTPBIN_OTT" ]]; then
+        logDebug "enrolling $HTTPBIN_OTT"
         # discard expected output that normally flows to stderr
         ENROLL_OUT="$(
-            ziti edge enroll "$httpbin_ott" 2>&1 \
+            ziti edge enroll "$HTTPBIN_OTT" 2>&1 \
                 | grep -vE '(generating.*key|enrolled\s+successfully)' \
                 || true
         )"
         if [[ -z "${ENROLL_OUT}" ]]; then
-            rm -f "$httpbin_ott"
-            logDebug "deleted $httpbin_ott after enrolling successfully"
+            rm -f "$HTTPBIN_OTT"
+            logDebug "deleted $HTTPBIN_OTT after enrolling successfully"
         else
             echo -e "ERROR: unexpected result during OpenZiti Identity enrollment\n"\
                     "${ENROLL_OUT}"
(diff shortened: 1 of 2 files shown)


sabedevops:
@qrkourier I could still use validation for the macOS and Windows execution paths

docusaurus/docs/learn/quickstarts/network/miniziti.bash
```diff
@@ -5,14 +5,17 @@ set -o nounset
...
+
     cat <<BANNER
                   __   ___
     |\/| | |\ | |  / |  |  |
     |  | | | \| | /_ |  |  |
 
 BANNER
 
-    if (( $# )); then
+    if [[ -n "$profile" ]] && [[ "$profile" != "miniziti" ]]; then
```
qrkourier: I have a slight preference for compound expressions over multiple invocations of `[[` or `test`.

```suggestion
    if [[ -n "$profile" && "$profile" != "miniziti" ]]; then
```

docusaurus/docs/learn/quickstarts/network/miniziti.bash
```diff
@@ -26,6 +29,9 @@ _usage(){
     echo -e "\n COMMANDS\n"\
             "   start\t\tstart miniziti (default)\n"\
             "   delete\t\tdelete miniziti\n"\
+            "   login\t\trun ziti edge login with miniziti context\n"\
+            "   show-admin\t\tprints admin user updb credentials\n"\
```
qrkourier: If you think that we'll want to expand on the list of potential objects for the verb `show`, then we should go ahead and define the API like `miniziti verb object`, e.g., `show admin` rather than separate couplets for each permutation.

On the other hand, if it's truly a one-off, then introducing a single couplet `show-admin` might be the only thing that makes sense, as written, for the sake of simplicity in the implementation.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
```diff
@@ -293,26 +361,60 @@ main(){
...
-    # delete and exit if --delete
     (( DELETE_MINIZITI )) && {
         deleteMiniziti 10
+
+        if (( MINIZITI_MODIFY_HOSTS )) && grep -q "$MINIZITI_PROFILE_MARKER" "$HOSTS_FILE"; then
+            checkSudoRequired
+            cleanHosts
+        fi
+
+        if [[ -d "$profile_dir" ]]; then
```
qrkourier: It will help readability if we stick to some capitalization convention for var and function names. It doesn't have to be uppercase, just consistent. I'm open to separate conventions for heritable, overrideable, global, local, iterable, etc., but that feels like overkill.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
```diff
@@ -293,26 +361,60 @@ main(){
...
+            logWarn "Deleting miniziti profile directory: $profile_dir"
+            rm -rf  "$profile_dir"
+        fi
+
+        cert_file="$(find "$HOME/.config/ziti/certs" -maxdepth 1 -type f -name "miniziti-controller.$MINIKUBE_PROFILE.*" -print -quit)"
```
qrkourier: TODO: make sure the `find` on Darwin is findutils find not a BSD variant because it may not have the same options.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
```diff
@@ -555,24 +657,28 @@ main(){
     #
 
     logDebug "fetching admin password from k8s secret to log in to ziti mgmt"
-    kubectl get secrets "ziti-controller-admin-secret" \
-        --namespace "${ZITI_NAMESPACE}" \
-        --output go-template='{{index .data "admin-password" | base64decode }}' \
-        | xargs ziti edge login "miniziti-controller.${MINIZITI_INGRESS_ZONE}:443" \
-            --yes --username "admin" \
+            getAdminSecret | xargs ziti edge login "miniziti-controller.${MINIZITI_INGRESS_ZONE}:443" \
```
qrkourier, line 692: not `miniziti_login()`?

docusaurus/docs/learn/quickstarts/network/miniziti.bash
```diff
@@ -169,36 +189,60 @@ installHosts() {
...
 logDebug() {
-    (( $# )) || {
-        logError "logDebug() takes 1 or more args"
-        return 1
-    }
-    echo "DEBUG: $*" >&3
+    logger "$*" >&3
+}
+
+ziti_wrapper() {
```
qrkourier, line 263: Possibly a future feature for another PR, but it'd be convenient if this function performed a login when necessary. That way, it's never necessary to explicitly to know about or think about running the login command by itself, just the CLI wrapper. The CLI JSON file has a `loginTime` property for the profile, and it might be possible to discover when the token expires. For example,

```bash
# pseudo bash
ziti_wrapper(){
if login profile exists; then
    if loginTime > expiry; then
        miniziti_login()
    fi
else
    miniziti_login()
fi
ziti "$@" --cli-identity "$MINIKUBE_PROFILE"
}
```

qrkourier (changes requested):
I like the direction and the patterns. I know you're also working on doc.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: Can update.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: Do you forsee any other show verbs at this time?

qrkourier: I don't.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: No, because `miniziti_login()` presumes a pre-exisitng `ziti-cli.json`, and this does not.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: Agreed, and on several lines, I was tempted to add in the `jq` dependency. Any objection? We take out odd `awk` statements that way

qrkourier: I'm reluctant to introduce a dependency on `jq`, yes. Merely eliminating `awk` doesn't seem like a good enough reason. We're using `awk` in a portable way up to this point.

sabedevops: I don't want to deal with JSON outside of `jq` if I can help it, so can we leave this request for a future feature?

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: I'll capitalize variables in `main()` that are intended as `readonly` pseudo constants.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: Should be resolved.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: Should be resolved.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
qrkourier: Good call on that.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: Confirmed with multiple sources this should work on mac.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: Would `admin-creds` be better? Some alternative that avoids this pitfall?

qrkourier: I think it's fine.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
```diff
@@ -89,24 +95,35 @@ detectOs(){
     fi
 }
 
-getClientOttPath(){
-    local ID_PATH="/tmp/${MINIKUBE_PROFILE}-client.jwt"
-    if [[ $# -eq 1 ]]; then
-        case $1 in
-            macOS|Linux)     echo "$ID_PATH"
+makeMinizitiXDGStateDir() {
```
qrkourier: Slight preference for sticking to dromedary case instead of deviating to a hybrid with some uppercase.

```suggestion
makeMinizitiXdgStateDir() {
```

docusaurus/docs/learn/quickstarts/network/miniziti.bash
```diff
@@ -293,26 +361,60 @@ main(){
...
-            logError "--profile needs a profile name not starting with a hyphen"
-            _usage; exit 1
-        fi
-        # print the alternative profile name if not default "miniziti"
-        banner "$MINIKUBE_PROFILE"
+    : "${ZITI_NAMESPACE:=${MINIKUBE_PROFILE}}"
+
+    MINIZITI_PROFILE_MARKER="miniziti-controller.$MINIKUBE_PROFILE."
+    MINIZITI_INGRESS_ZONE="$MINIKUBE_PROFILE.ziti"
+    STATE_DIR="$(makeMinizitiXDGStateDir "$DETECTED_OS")"
```
qrkourier:
```suggestion
    STATE_DIR="$(makeMinizitiXdgStateDir "$DETECTED_OS")"
```

qrkourier (changes requested):

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: I've simply removed `XDG` from the name because I find that variant repulsive. There is actually no consensus on that subject anywhere I'm aware of. Some projects encode this in style guides. In the future, changes like these should be the cause of churn only if we can point to a specific guideline infringement. Even better, a linter in a pipeline should enforce such conventions.

qrkourier:
> @qrkourier I could still use validation for the macOS and Windows execution paths

Looking good in WSL. Noticed a couple of things. Let's think these over to see if they deserve any adjustments.

1. `miniziti-controller.miniziti.ziti` isn't very aesthetically pleasing because it's redundant, but I see the good reasons it's built that way. How about `ziti-controller.miniziti`, i.e., the profile name becomes the TLD which only existed in the first place so that each Ingress host would have at least one `.` stop char.
2. Every time we change the domain name of the ingresses, users are forced to adjust their hosts db. It would be kind of us to slow down or eliminate forcing/foisting that change. Good reason to nail this round of changes so we're happy with it for a while.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
```diff
@@ -89,24 +95,35 @@ detectOs(){
...
-            Windows)         wslpath -w "$ID_PATH"
+        Linux)
+            state_dir="${XDG_STATE_DIR:-$HOME/.local/state}/miniziti"
             ;;
-        esac
-    else
-        logError "getClientOttPath() takes one param, DETECTED_OS"
-        return 1
+        Windows)
+            wsl_state_dir="${XDG_STATE_DIR:-$HOME/.local/state}/miniziti"
```
qrkourier:
```suggestion
            state_dir="${XDG_STATE_DIR:-$HOME/.local/state}/miniziti"
```

docusaurus/docs/learn/quickstarts/network/miniziti.bash
```diff
@@ -89,24 +95,35 @@ detectOs(){
...
+            state_dir="$(wslpath -w "$wsl_state_dir")"
```
qrkourier:
```suggestion
            windows_state_dir="$(wslpath -w "$state_dir")"
```

We're not using `windows_state_dir` yet, but we might if we decide to resurrect `getClientOttPath()`.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
```diff
@@ -638,139 +744,143 @@ main(){
...
-'INFO: Your OpenZiti Console is here:\thttps://miniziti-console.'"${MINIZITI_INGRESS_ZONE}"'\n'\
-'INFO: The password for \"admin\" is:\t"}}{{index .data "admin-password" | base64decode }}'\
-'{{"\n\n"}}'
+    echo -e "\n\n"
+    logInfo "Your OpenZiti Console is here: https://miniziti-console.${MINIZITI_INGRESS_ZONE}"
+    showAdminCreds
+    echo -e "\n\n"
 
-    logInfo "Success! Remember to add your edge client identity '$(getClientOttPath "${DETECTED_OS}")' in your tunneler, e.g. Ziti Desktop Edge."
+    logInfo "Success! Remember to add your edge client identity '$CLIENT_OTT' in your tunneler, e.g. Ziti Desktop Edge."
```
qrkourier: This doesn't quite work the same in WSL because `getClientOttPath()` would print the resolved Windows UNC path by which the user could load the OTT in Ziti Desktop Edge for Windows. Now it prints the Linux path, which isn't useful for the Windows user that wants to connect to Ziti services provided by miniziti.

The vars built from the state dir are used at least two ways: one way provides paths for things like the `ziti edge create router` command writing the OTT file, and the other for Windows users to find the client OTT with a UNC path that was written to a POSIX path in WSL.

One way to handle this is to resurrect and modify `getClientOttPath()` and call it in this line. It could take the value of $CLIENT_OTT as input and return the same value if macOS or Linux and otherwise the resolved Windows UNC path.

qrkourier (changes requested):

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: Good catch. I approached this in a slightly different, more generic way than your recommendation. Please re-test, and let me know if I got it right.

sabedevops:
> > @qrkourier I could still use validation for the macOS and Windows execution paths
> 
> Looking good in WSL. Noticed a couple of things. Let's think these over to see if they deserve any adjustments.
> 
>     1. `miniziti-controller.miniziti.ziti` isn't very aesthetically pleasing because it's redundant, but I see the good reasons it's built that way. How about `ziti-controller.miniziti`, i.e., the profile name becomes the TLD which only existed in the first place so that each Ingress host would have at least one `.` stop char.
> 
>     2. Every time we change the domain name of the ingresses, users are forced to adjust their hosts db. It would be kind of us to slow down or eliminate forcing/foisting that change. Good reason to nail this round of changes so we're happy with it for a while.

I wouldn't agree that this:

* miniziti-controller.miniziti
* miniziti-controller.test
* miniziti-controller.edgedev

Looks more aesthetically pleasing than this:

* miniziti-controller.miniziti.ziti
* miniziti-controller.test.ziti
* miniziti-controller.edgedev.ziti

I think the `.ziti` at the end gives a much nicer effect.

The question of `miniziti-controller` vs `ziti-controller` and the like should've been addressed in a prior PR where this convention was adopted. Addressing it here places unnecessary churn on this PR. For users using `--modify-hosts`, this is somewhat obviated in that such entries are cleaned up in both cases, (in this case, backward compatible) if in the future we decided to change this.

Edit:
A further point is that when `--no-hosts` is used, you end up with consistent entries in the following form:

`miniziti-[component].[ip_address].sslip.io`. 

Here, `sslip.io` is consistently appended. The way `.ziti` parrots this and distinguishes local entries from those of `sslip.io` is also visually nice. 

Compare:
```
192.168.50.252 miniziti-controller.miniziti.ziti miniziti-router.miniziti.ziti miniziti-console.miniziti.ziti
192.168.43.252 miniziti-controller.test.ziti miniziti-router.test.ziti miniziti-console.test.ziti
192.168.39.251 miniziti-controller.192.168.39.251.sslip.io miniziti-router.192.168.39.251.sslip.io miniziti-console.192.168.39.251.sslip.io
```

vs 

```
192.168.50.252 miniziti-controller.miniziti miniziti-router.miniziti miniziti-console.miniziti
192.168.43.252 miniziti-controller.test miniziti-router.test miniziti-console.test
192.168.39.251 miniziti-controller.192.168.39.251.sslip.io miniziti-router.192.168.39.251.sslip.io miniziti-console.192.168.39.251.sslip.io
```

docusaurus/docs/learn/quickstarts/network/miniziti.bash
qrkourier: Looks like it will work. I'll run in WSL this morning.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
```diff
@@ -169,36 +191,60 @@ installHosts() {
...
         return 1
-    }
-    echo "INFO: $*"
+    fi
+
+    caller_level="${caller##log}"
+    if (( MINIZITI_DEBUG )); then
+        echo -e "${caller_level^^} ${FUNCNAME[2]}:${BASH_LINENO[1]}: $*"
+    else
+        echo -e "${caller_level^^} $*"
```
qrkourier: interprets backslashes emitted by `pathToNative()` e.g.,

```bash
kbingham@oryp3win10:/tmp/tmp.BVnGn9z6ho$ bash miniziti.bash start
                  __   ___
    |\/| | |\ | |  / |  |  |
    |  | | | \| | /_ |  |  |

INFO waiting for minikube to be ready
INFO waiting for ingress-nginx to be ready
INFO upgrading openziti controller
INFO waiting for ziti-controller-cert-manager to be ready
INFO waiting for trust-manager to be ready
INFO waiting for ziti-controller to be ready
INFO cluster dns test succeeded
INFO Setting default ziti identity to: miniziti
INFO waiting for ziti-router to be ready
INFO miniziti-router is online
INFO waiting for ziti-console to be ready



INFO Your OpenZiti Console is here: https://miniziti-console.miniziti.ziti
INFO The password for 'admin' is: 'KsBMq26FzDjwP6YGHfPakf2DqfmL8rIx'



INFO Success! Remember to add your edge client identity '\wsl.localhost
                                                                       untu\home\kbingham\.local\state\miniziti\profiles\miniziti\identities\miniziti-client.jwt' in your tunneler, e.g. Ziti Desktop Edge.
```

docusaurus/docs/learn/quickstarts/network/miniziti.bash
```diff
@@ -89,24 +95,37 @@ detectOs(){
...
+        mkdir -p "$state_dir"
     fi
 
+    echo "$state_dir"
+}
+
+pathToNative() {
+    local path="$1"
+    case "$DETECTED_OS" in
+        Windows) wslpath -w "$path" ;;
```
qrkourier, line 129:
```suggestion
        Windows) wslpath -w "$path" | sed -E 's|\\|\\\\|g' ;; # escape backslashes in UNC path from interpretation by logger()
```

qrkourier (changes requested):

qrkourier:
Encountered a sporadic issue during WSL testing. It's as if the check for existing router succeeds incorrectly and so miniziti tries to create the already-created router. Not sure yet if related to changes in this branch. It seems to be occurring about 20-40% of the time, and `miniziti-router` is not being re-created because it always has the same Ziti ID.

```bash
kbingham@oryp3win10:/tmp/tmp.BVnGn9z6ho$ bash miniziti.bash ziti edge list edge-routers
╭────────────┬─────────────────┬────────┬───────────────┬──────┬────────────────╮
│ ID         │ NAME            │ ONLINE │ ALLOW TRANSIT │ COST │ ATTRIBUTES     │
├────────────┼─────────────────┼────────┼───────────────┼──────┼────────────────┤
│ dysMVtRKBq │ miniziti-router │ true   │ true          │    0 │ public-routers │
╰────────────┴─────────────────┴────────┴───────────────┴──────┴────────────────╯
results: 1-1 of 1

kbingham@oryp3win10:/tmp/tmp.BVnGn9z6ho$ bash miniziti.bash start
                  __   ___
    |\/| | |\ | |  / |  |  |
    |  | | | \| | /_ |  |  |

INFO waiting for minikube to be ready
INFO waiting for ingress-nginx to be ready
INFO upgrading openziti controller
INFO waiting for ziti-controller-cert-manager to be ready
INFO waiting for trust-manager to be ready
INFO waiting for ziti-controller to be ready
INFO cluster dns test succeeded
INFO Setting default ziti identity to: miniziti
error: error creating edge-routers instance in Ziti Edge Controller at https://miniziti-controller.miniziti.ziti:443/edge/management/v1. Status code: 400 Bad Request, Server returned: {
    "error": {
        "cause": {
            "field": "name",
            "reason": "name is must be unique",
            "value": "miniziti-router"
        },
        "code": "COULD_NOT_VALIDATE",
        "message": "The supplied request contains an invalid document or no valid accept content were available, see cause",
        "requestId": "LwgQuyqTB"
    },
    "meta": {
        "apiEnrollmentVersion": "0.0.1",
        "apiVersion": "0.0.1"
    }
}
```

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: Can you try again with the new changes?

qrkourier: I'm getting this consistently with https://raw.githubusercontent.com/openziti/ziti-doc/feature/multi_profile_support/docusaurus/docs/learn/quickstarts/network/miniziti.bash

```bash
INFO cluster dns test succeeded
+ logDebug 'fetching admin password from k8s secret to log in to ziti mgmt'
+ logger 'fetching admin password from k8s secret to log in to ziti mgmt'
+ local caller=logDebug
+ ((  1 < 1  ))
+ local 'message=fetching admin password from k8s secret to log in to ziti mgmt'
+ [[ fetching admin password from k8s secret to log in to ziti mgmt =~ ^r'(.+)'$ ]]
+ caller_level=Debug
+ ((  MINIZITI_DEBUG  ))
+ line='DEBUG fetching admin password from k8s secret to log in to ziti mgmt'
+ [[ -n '' ]]
+ echo -e 'DEBUG fetching admin password from k8s secret to log in to ziti mgmt'
+ getAdminSecret
+ kubectl_wrapper get secrets ziti-controller-admin-secret --namespace miniziti --output 'go-template={{index .data "admin-password" | base64decode }}'
+ xargs ziti edge login miniziti-controller.miniziti.ziti:443 --yes --cli-identity miniziti --username admin --password
+ minikube kubectl --profile miniziti -- --context miniziti get secrets ziti-controller-admin-secret --namespace miniziti --output 'go-template={{index .data "admin-password" | base64decode }}'
The connection to the server miniziti-controller.miniziti.ziti:443 was refused - did you specify the right host or port?
```

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: This should be resolved.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: This should be resolved.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: This should also be resolved.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
```diff
@@ -617,160 +743,164 @@ main(){
...
+        logDebug "ignoring identity $HTTPBIN_NAME"
     fi
 
-    if  ! ziti edge list configs 'name="httpbin-intercept-config"' --csv \
+    if  ! ziti_wrapper edge list configs 'name="httpbin-intercept-config"' --csv \
         | grep -q "httpbin-intercept-config"; then
         logDebug "creating config httpbin-intercept-config"
-        ziti edge create config "httpbin-intercept-config" intercept.v1 \
+        ziti_wrapper edge create config "httpbin-intercept-config" intercept.v1 \
             '{"protocols":["tcp"],"addresses":["httpbin.'"${MINIZITI_INTERCEPT_ZONE}"'"], "portRanges":[{"low":80, "high":80}]}' >&3
```
qrkourier, line 821: This results in colliding intercepts when miniziti profiles are running concurrently. Consider building the intercept zone with the profile name.

qrkourier (changes requested):

qrkourier:
macOS test result

```bash
ec2-user@ip-172-31-28-110 tmp.0961Qy4q % curl -O https://raw.githubusercontent.com/openziti/ziti-doc/bf189ce7a950d640bbaabfa6ff69c4c16c7cd2a3/docusaurus/docs/learn/quickstarts/network/miniziti.bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 33399  100 33399    0     0   799k      0 --:--:-- --:--:-- --:--:--  959k

ec2-user@ip-172-31-28-110 tmp.0961Qy4q % ll
total 72
-rw-r--r--  1 ec2-user  staff    33K Sep  7 17:19 miniziti.bash

ec2-user@ip-172-31-28-110 tmp.0961Qy4q % bash miniziti.bash start
miniziti.bash: line 227: MINIZITI_DEBUG: unbound variable

ec2-user@ip-172-31-28-110 tmp.0961Qy4q % bash -x miniziti.bash start
+ set -o errexit
+ set -o nounset
+ set -o pipefail
+ HOSTS_FILE=/etc/hosts
+ main start
+ BINS=(ziti minikube kubectl helm sed)
+ declare -a BINS
+ for BIN in '"${BINS[@]}"'
+ command -v ziti
+ logError 'this script requires commands '\''ziti minikube kubectl helm sed'\''. Please install on the search PATH and try again.'
+ logger 'this script requires commands '\''ziti minikube kubectl helm sed'\''. Please install on the search PATH and try again.'
+ local caller=logError
+ ((  1 < 1  ))
+ local 'message=this script requires commands '\''ziti minikube kubectl helm sed'\''. Please install on the search PATH and try again.'
+ [[ this script requires commands 'ziti minikube kubectl helm sed'. Please install on the search PATH and try again. =~ ^r'(.+)'$ ]]
+ caller_level=Error
+ ((  MINIZITI_DEBUG  ))
miniziti.bash: line 227: MINIZITI_DEBUG: unbound variable
```

docusaurus/docs/learn/quickstarts/network/miniziti.bash
sabedevops: This should be resolved.

sabedevops: This should be resolved.

qrkourier (approved):
:ship:

PR #674, opened 2023-09-08
run ziti cli inside controller container

qrkourier:


docusaurus/docs/learn/quickstarts/network/miniziti.bash
```diff
@@ -224,7 +224,7 @@ logger() {
         return 1
     fi
 
-    local message="$*"
```
qrkourier: oops

dovholuknf (approved):

PR #679, opened 2023-09-08
Breakfix/miniziti refactor issues

sabedevops:


sabedevops:
@dovholuknf && @qrkourier This restores miniziti functionality. Can we please get this merged? There are a few, minor outstanding issues I will address through this branch later, but for now, this gets things back to sanity and working.

docusaurus/docs/learn/quickstarts/network/miniziti.bash
```diff
@@ -471,7 +473,7 @@ main(){
 
         if check_command ziti &>/dev/null; then
             logWarn "Removing $MINIKUBE_PROFILE profile identity from ziti-cli.json"
-            ziti edge logout --cli-identity "$MINIKUBE_PROFILE" >&3
+            #ziti edge logout --cli-identity "$MINIKUBE_PROFILE" >&3
```
qrkourier: If this is intentionally commented, then why not also remove the enclosing check and log message?

qrkourier (changes requested):
Should we double-check the output token files to ensure they resemble a JWT? At a minimum, a `test -s`, but perhaps a minimum byte count?

sabedevops:
> Should we double-check the output token files to ensure they resemble a JWT? At a minimum, a `test -s`, but perhaps a minimum byte count?

I've added `--exit-status` to `jq` instead.

qrkourier (approved):

PR #680, opened 2023-09-11
miniziti configurable timeout

qrkourier:


sabedevops:
If you're intentionally normalizing all wait periods to a single global value, then this PR looks good to me.

dovholuknf (approved):

PR #706, opened 2023-09-21
Writeup on how to renew certs

gberl002:


dovholuknf:
image

There's already a "PKI Troubleshooting" section. This should be nested under that section or it should be added to that doc instead...

Actually, this seems like two pages in one. There is troubleshooting stuff in there along with "how to" stuff. I think it would make sense to spearate this into the two parts

dovholuknf:
~~feels more like it should be reference or a guide in the current form. nowhere in there do i see "here is your problem, do this to fix it"~~

Nevermind, i see it's a guide. somehow I didn't notice! :/

dovholuknf:
I can't find the preview of this in the LHN, can you tell me how to find this page?

dovholuknf:
> I can't find the preview of this in the LHN, can you tell me how to find this page?

I see why, you haven't merged from main. It has a fix to make the top nav links, link into the vercel preview. It was sending me to the 'main' doc site. I've since figured that out and found the page

dovholuknf (approved):

PR #725, opened 2023-10-02
add a sed to the landing

### LINKED ISSUE #719: place correct nav links in Docusaurus build output files
Improve gendoc.sh to replace:

```
<a href="https://landing.openziti.io/" class="custom-logo-link"...
```

with this during a release:

```
<a href="https://openziti.io/" class="custom-logo-link"...
```

and this during a PR build:

```
<a href="https://{{vercel preview domain}}//" class="custom-logo-link"...
```

dovholuknf:
trying to make the previews work properly and not always point to the 'official' doc

closes #719


gendoc.sh
```diff
@@ -186,6 +186,9 @@ if test -e "${script_root}/docusaurus/build/landing.html"; then
   echo "landing.html detected. overwriting index.html with landing.html"
   cp "${script_root}/docusaurus/build/index.html" "${script_root}/docusaurus/build/index.original.html"
   cp "${script_root}/docusaurus/build/landing.html" "${script_root}/docusaurus/build/index.html"
+  echo sed -i 's/<.*class="custom-logo-link"/<a href=\"\/\" class="custom-logo-link"/g' "${script_root}/docusaurus/build/index.html"
+  sed -i 's/<.*class="custom-logo-link"/<a href=\"\/\" class="custom-logo-link"/g' "${script_root}/docusaurus/build/index.html"
```
qrkourier: `-i ''` for Darwin compatibility, less greedy regex in case the generated HTML wraps, and `|` to eliminate escaping the backslash

```suggestion
  sed -Ei '' 's|(<a href=")(.*)(" class="custom-logo-link")|\1/\3|g' "${script_root}/docusaurus/build/index.html"
```

dovholuknf: funny enough i didn't see your `|` comment but I had already moved over to that for legibility and forgot to update this one! too funny. i'll give it a try on macos and see what happens and fix the first sed

qrkourier: This is even less greedy because it won't match two elements on the same line that both happen to have the same class.

```bash
sed -Ei '' 's|(<a href=")([^"]+)(" class="custom-logo-link")|\1/\3|g' "${script_root}/docusaurus/build/index.html"
```

dovholuknf: right. i understand your comment but if that situation ever actually happens, i'll be happy to come here and update this script myself. until then, it's not necessary to make this sed any more complicated than it is.

dovholuknf: after more reflection, I think the simplified version works well. worked on linux and macos and the vercel-deployed preview works properly so i'm calling it "gtg"! :)

qrkourier (approved):

PR #727, opened 2023-10-03
add an if on our jobs so they don't run on forks by default

dovholuknf:


qrkourier (approved):
It's a non-issue for most forks because forks don't have Actions workflows enabled by default.

PR #732, opened 2023-10-08
Fix the "Quickstart Guides" in Introduction page.

cinar:
The link for `Quickstart Guides` is currently pointing at `/docs/learn/quickstarts/network` which does not exist. The small change here is fixing that link by pointing it to `/docs/category/network` instead. This is aligned with the navigation links.


cinar:
Based on @dovholuknf's comment on [Fix the "Quickstart Guides" in Introduction page. #2](https://github.com/openziti/openziti.github.io/pull/2#issuecomment-1743386446). Moving the change to this repository instead.

dovholuknf (approved):
thanks for pointing this out

PR #803, opened 2024-02-05
document creating a database snapshot

qrkourier:


docusaurus/docs/guides/database-lifecycle.mdx
```diff
@@ -0,0 +1,44 @@
+---
+title: Database Lifecycle Management
+sidebar_position: 30
+---
+
```
plorenz: I'd maybe link to bbolt, since that's what we actually use. They are compatible, so I think we can still say it's Bolt.
https://github.com/etcd-io/bbolt

docusaurus/docs/learn/quickstarts/network/help/upgrade-quickstart-network.md
```diff
@@ -17,7 +17,7 @@ While you can certainly back up every file relating to your network, there are o
 back up the necessary files follow these steps:
 
 1. Backup the database:
-    1. Run the following
+    1. Run the following to create a database snapshot via the REST API.
```
dovholuknf: this example shows using the ziti CLI. i'd argue that you should just revert this change

docusaurus/docs/guides/database-lifecycle.mdx
```diff
@@ -0,0 +1,44 @@
...
+The OpenZiti Controller uses a [Bolt](https://github.com/boltdb/bolt) database to store its state. The database is a writeable file in a path specified by the controller's configuration property `.db` ([link to reference](/reference/30-configuration/controller.md#db)).
+
+Use the controller's built-in database snapshot operation to preserve data integrity. The snapshot operation creates a snapshot file in the same filesystem directory as the main database file on the controller host.
+
+There are three ways to invoke the snapshot operation.
+
+1. Agent (IPC): run the `snapshot-db` command on the controller host.
```
dovholuknf: IMO, ziti would only have one way of donig this. i'd like to only see "use the ziti cli" here. I don't think it's necessary to add in the complexity of the agent here.) I feel like the agent method should only be a stopgap for when the system isn't responding to the rest API.

qrkourier: I've trimmed it down to:

1. login then snapshot with CLI anywhere REST API is reachable
2. snapshot on controller host without login

It's unnecessary to redundantly document the REST operation because it's adequately covered by the spec and generated reference, which is now linked instead of explained.

qrkourier:
@plorenz Should we mention which POSIX perms are required to send a command to the IPC agent? ~~I didn't see any Unix domain sockets for the controller process, so I'm curious how that works.~~ I see `/tmp/gops-agent.$(pgrep -x ziti).sock` which has filemode allowing the controller's EUID to write to the Unix domain socket.

plorenz:
> @plorenz Should we mention which POSIX perms are required to send a command to the IPC agent? ~I didn't see any Unix domain sockets for the controller process, so I'm curious how that works.~ I see `/tmp/gops-agent.$(pgrep -x ziti).sock` which has filemode allowing the controller's EUID to write to the Unix domain socket.

Sure. It's definitely confused people before, when the controller was running as a different user and one needed to use sudo to access via IPC.

qrkourier:
I'll change this guide to focus on backing up the controller with the `ziti edge db snapshot` (via REST API) command, and to remove the mentions of the IPC method, because it's too confusing to promote both alternatives.

qrkourier:
@dovholuknf I completed the change we discussed to focus this short guide on backing up the controller db with the edge snapshot operation.

dovholuknf (approved):

PR #810, opened 2024-02-27
pages/appetizer: fix missing command in codeToCopy

ruzko:
The code to copy doesn't work, because the user needs to enter the directory of the cloned repo before go can see what it should build


dovholuknf (approved):
oh my! thanks! i dunno how i missed that

dovholuknf:
oh no -- can you amend your commit with a signed commit? we require commits be signed:
> The base branch requires all commits to be signed.

see the check that failed

ruzko:
There, it was a bit of work since I didn't have signing set up; but now I'm all set for secure, auditable commits in the future :)

PR #816, opened 2024-03-22
update router deploy ref

qrkourier:
align router deploy doc with Linux package and Docker norms


qrkourier:
aggregating changes in #820

PR #817, opened 2024-03-25
organize deployments refs as guides

qrkourier:


qrkourier:
aggregating changes in #820

PR #814, opened 2024-03-14
update controller deploy ref

qrkourier:

PR #820, opened 2024-03-28
organize and add ops guides

qrkourier:


docusaurus/docs/guides/deployments/10-controller/10-deploy.mdx
```diff
@@ -1,45 +1,42 @@
...
+| sudo bash -s openziti-controller
+```
 
-At a minimum, you must set these to start the service with bootstrapping enabled (the default).
+You must provide these two answers to start the service.
 
 1. In `/opt/openziti/etc/controller/env` (the systemd env file for the controller service)
     1. Set `ZITI_CTRL_ADVERTISED_ADDRESS` to the FQDN of the controller
-    1. Temporarily set `ZITI_PWD` to the desired management password for user `admin`
+    1. Temporarily set `ZITI_PWD` to the desired management password for user `admin`. You may delete this password after bootstrapping for security.
```
dovholuknf: this shouldn't be optional. it should be mandatory. i'm still not sold on the whole bootstrapping of the environment for the user. i still dislike the idea, on the whole, for this usecase.

dovholuknf: I still keep coming back to a three step process:
* step 1 install the package
* step 2 bootstrap the package/install
* step 3 set it up to run it as a service

docusaurus/docs/guides/deployments/10-controller/15-transplant.mdx
```diff
@@ -0,0 +1,51 @@
+---
+title: Transplant a Controller Configuration
```
dovholuknf: Transplant a Controller Configuration -> Migrating an Existing Install

docusaurus/docs/guides/deployments/10-controller/10-deploy.mdx
qrkourier: > this shouldn't be optional. it should be mandatory. i'm still not sold on the whole bootstrapping of the environment for the user. i still dislike the idea, on the whole, for this usecase.

It is mandatory.

docusaurus/docs/reference/config-types/_category_.json
```diff
@@ -1,3 +1,3 @@
 {
-  "label": "Builtin Config Types"
+  "label": "Config Types"
```
dovholuknf: These really are the "Built-in Config Types". I don't think you should change this?

docusaurus/docs/guides/kubernetes/hosting/_category_.json
```diff
@@ -1,5 +1,5 @@
 {
-  "label": "Hosting OpenZiti",
+  "label": "Self Hosting",
```
dovholuknf: Isn't any "hosting" by definition "self"? Feels like it to me. I prefer the former wording

docusaurus/docusaurus.config.js
```diff
@@ -87,6 +87,10 @@ const config = {
           return undefined;
         },
         redirects: [
+          {
+            to: '/docs/learn/core-concepts/data-flow-explainer',
```
dovholuknf: is this change in a different pr already? i'd approve this on a separate pr. seems reasonable to move this to where you put it

docusaurus/docs/guides/index.md
```diff
@@ -3,9 +3,7 @@ title: Overview
 sidebar_position: 10
 ---
 
-## What is a Guide?
-
-A guide is like a recipe. It's a type of documentation that is more advanced and solution-focused than other types. A guide is prescriptive and procedural. For example, "How to deploy a Ziti controller with Helm."
+This section contains operational guidance for specific tasks such as deploying a controller with Helm or as a Linux system service, backing up the controller's database, and migrating controller or router configurations.
```
dovholuknf: I'd like to keep the "what is a guide" with definition. it sets the stage for the type of documentation located in this section. I could see these new words going along with the former words as they build on the example from the old wording

docusaurus/docs/guides/hsm/index.md
```diff
@@ -1,10 +1,8 @@
 ---
-sidebar_position: 3
 id: hsmIndex
+title: Hardware Security Modules
```
dovholuknf: PKCS11 is relevant here as it is an API to interact with HSMs. I don't think removing it is necessary

docusaurus/docs/reference/config-types/_category_.json
qrkourier: The "built-in" qualifier is preserved in the title, in case there's any possibility the reader is confused. This changes the sidebar label to be shorter because it was unnecessarily long and unnecessarily qualified for a sidebar label.

docusaurus/docusaurus.config.js
qrkourier: No, everything is aggregated in this PR only.

qrkourier:
Copying feedback from https://github.com/openziti/ziti-doc/pull/814 because aggregating changes in this branch

docusaurus/docs/guides/kubernetes/hosting/_category_.json
qrkourier: Yeah, I don't like "Self Hosting" either. I was trying to distinguish between the two types of info we have about K8s: "deployments" and "tunneling" without further overloading the term "hosting." I'll try those as section labels.

docusaurus/docs/guides/index.md
qrkourier: Fine by me. It struck me as superfluous but it might be helpful.

docusaurus/docs/guides/hsm/index.md
qrkourier: Ah, I may have misunderstood the intent of the article. We're specifically documenting our support of PKCS#11, not HSMs in general, and not other HSM interfaces like PARSEC, right?

dovholuknf: The page references and uses pkcs11. i just don't feel the need to remove it from the title and i'm trying to reduce changes that aren't related to "running ziti in production". i'd like to keep changes like this outside of the scope of this pr

gendoc.sh
```diff
@@ -68,14 +67,14 @@ if [[ "${SKIP_GIT}" == no ]]; then
...
   git config --global --add safe.directory $(pwd)
-  clone_or_pull "https://github.com/openziti/ziti" "ziti-cmd"
-  clone_or_pull "https://github.com/openziti/ziti-sdk-csharp" "ziti-sdk-csharp"
-  clone_or_pull "https://github.com/openziti/ziti-sdk-c" "ziti-sdk-c"
-  clone_or_pull "https://github.com/openziti/ziti-android-app" "ziti-android-app"
-  clone_or_pull "https://github.com/openziti/ziti-sdk-swift" "ziti-sdk-swift"
-  clone_or_pull "https://github.com/openziti/ziti-tunnel-sdk-c" "ziti-tunnel-sdk-c"
-  clone_or_pull "https://github.com/openziti/helm-charts" "helm-charts"
-  clone_or_pull "https://github.com/openziti-test-kitchen/kubeztl" "kubeztl"
+  clone_or_pull "https://github.com/openziti/ziti" "ziti-cmd" release-next >/dev/null
```
dovholuknf: why?

qrkourier: Override the branch `main`. The new README files for the container images only exist in release-next.

docusaurus/docs/guides/05-troubleshooting/_category_.yml
qrkourier, line 1: this sidebar category was moved, not deleted

docusaurus/docs/guides/Public_Cloud_Deployment/README.md
qrkourier, line 1: moved, not deleted

docusaurus/docs/guides/database-backup.mdx
qrkourier, line 1: moved, not deleted

docusaurus/docs/guides/deployments/10-linux/20-router/_category_.yml
```diff
@@ -1,4 +1,4 @@
 label: Router
 link:
   type: doc
-  id: deployment
+  id: deploy
```
qrkourier: consistency

docusaurus/docs/guides/kubernetes/_category_.yml
qrkourier, line 1: moved

docusaurus/docs/guides/local-gateway/_category_.yml
qrkourier, line 1: moved

docusaurus/docs/guides/securing-apis/_category_.yml
qrkourier, line 1: moved

docusaurus/docs/learn/core-concepts/index.md
qrkourier, line 1: moved this content to sidebar category title, description props

docusaurus/docs/reference/deployments/01-controller.md
qrkourier, line 1: moved

docusaurus/docs/reference/deployments/02-router/01-deployment.md
qrkourier, line 1: moved

docusaurus/docs/reference/tunnelers/60-linux/10-debian-package.mdx
qrkourier, line 1: the linux tunneler article is broken up based on feedback that it was too long and busy into a category of more focused articles

docusaurus/docs/reference/tunnelers/80-kubernetes/_category_.yml
qrkourier, line 1: the kubernetes tunneling articles were reorganized here from guides

docusaurus/docs/reference/tunnelers/linux/container/_category_.yml
qrkourier, line 1: moved

docusaurus/docs/reference/tunnelers/linux/index.md
qrkourier, line 1: moved

gendoc.sh
qrkourier: reverted branch override now that the files are available in `main`

docusaurus/docs/_cli-log-levels-and-formats.md
```diff
@@ -2,7 +2,7 @@
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-All Ziti components log to standard output and standard error file descriptors.
+All Go Ziti components log to standard output and standard error file descriptors, so their output appears in the systemd journal (`journalctl -xeu ziti-controller.service`) when installed from a Linux package, and the Docker log (`docker compose logs ziti-router`) when run as a container. The C/C++ `ziti-edge-tunnel` has distinct log levels and controls, described in [Linux tunneler troubleshooting](/reference/tunnelers/60-linux/50-linux-tunnel-troubleshooting.mdx).
```
dovholuknf: I'd like to remove "Go ". It doesn't add any usuable information to the reader to know that these are "go" binaries and is something that might end up being rendered obsolete someday in the future which makes me dislike such a seemingly small addition. I'd like to return to the original wording. 

The remaining information was once 'reference' doc but here you've moved it to the guides section. This makes it feel wrong when reading it. It no longer feels relevant to logging in linux.

This section should focus on:
* You will log with json
* You will log to journald
* here's how you ship logs from the controller to some log collection agent

This is a guide. it should tell you how to do it and not present different options

docusaurus/docs/_cli-log-levels-and-formats.md
```diff
@@ -13,17 +13,17 @@ Output from Ziti components comes in three distinct styles. Choose the style of
 - `json` - a machine-readable format targeting automated processes for log aggregation/searching
 - `text` - a human-readable format using plain text (no ascii escape codes)
 
-```text
-ziti-router run ./router.yml --log-formatter pfxlog
+```text title="Run the router without colorized log levels"
+ziti router run config.yml --log-formatter text
```
dovholuknf: why does "router" show up under "controller" -- this doesn't fit here either. i think the whole section probably needs to be rearranged per the prior comment

dovholuknf: Now I see it was reused in two places... I think what this turns into is that it just doesn't make sense to reuse this content...

docusaurus/docs/_linux-package-repo.mdx
```diff
@@ -0,0 +1,45 @@
+
+import Details from '@theme/MDXComponents/Details';
+
+<Details>
+<summary>Manually set up DEB repository</summary>
```
dovholuknf: I think it's worthwhile to add Ubuntu/Fedora/RedHat onto these too. I think people will naturally search for "how to install openziti on fedora" not "by rpm". At least the most popular ones like Ubuntu/Fedora/RedHat?

docusaurus/docs/downloads.mdx
```diff
@@ -69,7 +69,9 @@ This app provides the Ziti tunneler background service (`ziti-edge-tunnel`) and
 <!-- begin super-tab Linux -->
 <TabItem value="Linux">
 
-Install the `ziti` command or the `ziti-edge-tunnel` tunneler daemon as a package or a binary.
+Install the `ziti` command or the `ziti-edge-tunnel` tunneler as a package or a binary.
+
+There are systemd service packages available for `ziti-controller.service` and `ziti-router.service` too.
```
dovholuknf: i see `ziti-controller.service` here but "The controller package `openziti-controller`" below. That seems like it should be the same and I personally think this should stay `ziti-controller` and `ziti-router`

qrkourier: It seems like it should be the same, but there's no easy way to make it that way. Do you want the longer answer?

docusaurus/docs/guides/05-troubleshooting/_category_.yml
dovholuknf, line 1: just checking to make sure any `_category_.yml` changes in this PR are now intentional? Why is this one being removed?

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+---
+
+import LinuxPackageRepo from '/docs/_linux-package-repo.mdx';
+import CliLogFormats from '/docs/_cli-log-levels-and-formats.md';
+
+This article is about deploying a Ziti Controller as a Linux system service. [The controller introduction](/learn/introduction/03-components.md#openziti-controller) introduces the architectural concepts used in this guide.
+
+## Installation
+
+The controller package `openziti-controller` provides a systemd service unit and optional bootstrapping. The package depends on the `openziti` package which provides the `ziti` CLI. The easiest way to install both is the cross-platform install script for RPM and Debian distributions.
```
dovholuknf: "cross-platform install script" somehow rings a bit strange for me. I usually think of cross-platform as being able to run on more than a single platform (linux). I know what you're shooting for here, but in the spirit of "less is more", it feels like this could just remove "cross-platform". Also I'd like to remove "easiest". This guide should just stick to the facts:
You can install the `openziti` package by executing the convenience script or by installing the package manually. something like that

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+```text
+curl -sS https://get.openziti.io/install.bash \
+| sudo bash -s openziti-controller
+```
+
+<LinuxPackageRepo />
+
```
dovholuknf: After installing manually, the user needs to be told 'how' to install it. I presume it's either:
```
apt install openziti
```
or
```
dnf install openziti
``` 
?

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
+---
+title: Controller Deployment
```
dovholuknf: After looking at this for a bit, it felt redundant to see "Deployments -> Linux -> Controller Deployment" and the sidebar also shows just "Controller". I think "Deployment" here is redundant and can/should be removed.

I'd also like to see the preamble here to indicate the three phases (are there more?) of deployment:
* Installation/Package installation
* Configuration
* Starting/Running as a service

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+The controller package `openziti-controller` provides a systemd service unit and optional bootstrapping. The package depends on the `openziti` package which provides the `ziti` CLI. The easiest way to install both is the cross-platform install script for RPM and Debian distributions.
+
+```text
+curl -sS https://get.openziti.io/install.bash \
+| sudo bash -s openziti-controller
+```
+
+<LinuxPackageRepo />
+
+## Bootstrapping
```
dovholuknf: I'd like to rework this to be "Configuration" instead of "Bootstrapping".  It should offer two sections
* New Configuration
* Migrating an Existing Configuration

New Configuration is where you can introduce the newly generated "bootstrapping" process but I feel strongly that this should be a very explicit step that the user must complete. One of the lessons I've learned from the quickstarts is that people will get this wrong. Making it an explicit step, not implicitly accomplished, puts the onus onto the user to follow the instructions and not get it wrong.

At this point in the reading, it's not clear enough to me from reading the Bootstrapping section what will happen. I'd like to see the expectations/flow documented here and I'd like to steer people to the smallest set of steps we can and tell them how to do it with a "do this" sort of flow.

This would then also obviate the "Migrate" subpage on this node too. (maybe... depends on how it ends up)

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+You can opt-in to generating a PKI, config, and database by installing interactively with prompts or by setting these values.
+
+1. In `/opt/openziti/etc/controller/service.env` (the answer file for the service)
```
dovholuknf: I'd like to remove any 'opt-in' stuff and tell them exactly what to set and how to execute it from a minimal perspective. We know there are really just one thing that one really MUST get right from the start, the advertised address/port for the controller (which must support IP-only deployments). For a new install we should just tell them this is vitally important because they are about to generate a PKI (by running some command/commands we provide for them to run).

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+    1. Set `ZITI_BOOTSTRAP=true`
+1. In `/opt/openziti/etc/controller/bootstrap.env` (the answer file for bootstrapping the controller)
+    1. Set `ZITI_CTRL_ADVERTISED_ADDRESS` to the FQDN of the controller
+    1. Set `ZITI_PWD` to the desired management password for user `admin`. You may delete this after bootstrapping.
+
+### Generating a Public Key Infrastructure
```
dovholuknf: This would already be accomplished based on the prior comment. This is important, and they need to understand that this all exists, but it's in the scope of backing up this PKI by offlining the root key(s) and when and how to rotate important certs. Any optional type stuff should be reseved for a reference type document. Let's tell them the 95% install case here. The `ZITI_BOOTSTRAP_PKI` stuff should be removed. Instead, we should just tell them here's what we're doing and why.

Rotating the controller's server cert on the user's behalf should be just a presumption that happens on some interval and this should just tell them that it's going to happen and on what interval to expect it to happen.

Using `ALT SERVER CERTS` should be mentioned in here, particularly in relation to zrok/browzer.

Does the signing intermediate need to get rotated? How long will it be valid for? That's should all be captured here for people to be aware of.

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+The controller service will generate a root CA and issue an intermediate CA cert during the first startup.
```
dovholuknf: As stated before, i feel strongly that the configuration step should be separated from the startup step.

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+The controller needs a CA to issue certificates for edge identities during enrollment. This CA is known as the edge signer. It's a good idea to secure the root in a different location and use an intermediate CA for this purpose. This allows you to recover in a scenario where the intermediate was lost without changing the root of trust and invalidating all edge enrollments.
```
dovholuknf: Tell the user how to secure the root in a different location. Like, they need to know where is "the root" and then they need to be told to verify it is safe, and then `rm` it

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+Disable bootstrapping a PKI by setting `ZITI_BOOTSTRAP_PKI=false` in `/opt/openziti/etc/controller/service.env`.
+
+Check out [the PKI page](/learn/core-concepts/pki.md) for an overview of these concepts.
+
+### Generating a Configuration
+
+The controller service will generate a configuration during the first startup.
```
dovholuknf: Implementation question: Can we make the existence of the answer file a requirement here? As in, the user must have placed an answer file into the expected location for the service to startup the first time, so that the proper configuration can be generated? That way the configuration step mentioned before is basically "make the answer file" (exactly how the answer file is created is an implementation detail that doesn't need to be documented). Then the process could remove the answer file for the user so that it's not left behind. That would allow the ZITI_PWD to be put into the answer file as well since the password gets removed by the initialization step.

Then when the controller starts it can verify the answer file exists (or doesn't) to use that to determine if the config should be generated(regenerated) on start.

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+The controller's configuration is loaded from a YAML file ([reference](/reference/30-configuration/controller.md)). The Linux system service will generate a valid configuration during the first startup unless one already exists.
```
dovholuknf: Some, if not all of this, feels redundant to "The controller service will generate a configuration during the first startup." I also think it's not necessary to tell people it's loaded from a YAML file and just reference "the controller's configuration file".

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+Check out [the PKI page](/learn/core-concepts/pki.md) for an overview of these concepts.
+
+### Generating a Configuration
+
+The controller service will generate a configuration during the first startup.
+
+The controller's configuration is loaded from a YAML file ([reference](/reference/30-configuration/controller.md)). The Linux system service will generate a valid configuration during the first startup unless one already exists. 
+
+The filename of the configuration file, relative to the Linux service's working directory (`/var/lib/ziti-controller`), is specified in `/lib/systemd/system/ziti-controller.service` as an argument to the `ExecStart` command. You may override this in `/etc/systemd/system/ziti-controller.service.d/override.conf`.
```
dovholuknf: why allow the user to override this file location? it just adds more stuff to document and more permutations to test. Is this strictly necessary for some reason? Does the default location change? Can we tell them what the default location is without making them crack open the unit file to find it?

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+Disable bootstrapping a configuration by setting `ZITI_BOOTSTRAP_CONFIG=false` in `/opt/openziti/etc/controller/service.env`.
+
+### Bootstrapping the Database
+
+The controller service will bootstrap a database during the first startup.
+
+The controller requires a BoltDB database to store its state. The Linux system service will initialize a database with a default admin password during the first startup unless one already exists.
```
dovholuknf: "The controller requires a BoltDB database to store its state."  feels like the lead sentence to me. The previous sentence seems redundant to this one.

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+You must specify the management password for the default admin user before starting the service. This is done by setting `ZITI_PWD` in `/opt/openziti/etc/controller/bootstrap.env` or in the root-only file specified by `LoadCredential` in `/etc/systemd/system/ziti-controller.service.d/override.conf`. You may delete the password after bootstrapping for security.
```
dovholuknf: this fits in with the prior comment wrt implementation. the process should just remove the answer file for you imo and this section would just spell out that the file is removed after startup. Then "You may delete the password after bootstrapping for security." becomes "The answer file with the password is removed for you on successul configuration for security." (or something like that)

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+Disable bootstrapping the database by setting `ZITI_BOOTSTRAP_DATABASE=false` in `/opt/openziti/etc/controller/service.env`.
+
+## Firewall
+
+The controller listens on a single configurable TCP port: `1280/tcp`. This TLS server employs SNI to select the correct certificate for presentation when there are multiple certificates. Ziti clients use ALPN to negotiate a connection to the router control plane (`ziti-ctrl`) or the REST APIs (`h2`, `http/1.1`).
```
dovholuknf: while good information, this is all reference doc and doesn't really help the user with installing the controller (and isn't relevant to the firewall). This should be removed or a link to the refernce doc added:
```
This TLS server employs SNI to select the correct certificate for presentation when there are multiple certificates. Ziti clients use ALPN to negotiate a connection to the router control plane (`ziti-ctrl`) or the REST APIs (`h2`, `http/1.1`).
```

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+You may set `ZITI_CTRL_ADVERTISED_PORT` in `/opt/openziti/etc/controller/bootstrap.env` to bootstrap with a different port.
```
dovholuknf: This is configuration-related and belongs in the "how to make your configuration" section. What I expected to see here is information related to firewalld or iptables or other things like that. I also expected to see some sort of way to probe your system after starting the controller to ensure that it's able to be connected to properly. That might be where you bring the ALPN stuff into the picture since to probe the port properly will need ALPN to be specified

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+Clients "learn" the controller's address and port when they enroll, so it is necessary to re-enroll or replace the identity if the controller's advertised address or port changes.
```
dovholuknf: If you want to keep this, this also is more about configuration and "getting it right". It's definitely not related to "firewall" .

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
dovholuknf: After rereading this, it does feel relevant because they need to determine what port the controller **is** listening on, so that is definitely relevant to firewall... but it's in the scope of "go find out what port the controller is running on", not "you can configure it". Reading the config file is one way (maybe the best way), or using `ss` or something else.

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+## Agent
+
+The controller provides an IPC agent for administration. The agent listens on a Unix domain socket inside the filesystem namespace of the controller service. Here's an example for querying the controller's agent for statistics.
```
dovholuknf: This feels like debugging/troubleshooting tpye of information and doesn't feel relevant to "installation" to me. Feels out of place?

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+```buttonless title="Output"
+goroutines: 50
+OS threads: 22
+GOMAXPROCS: 16
+num CPU: 16
+```
+
+## Logging
+
+<CliLogFormats/>
```
dovholuknf: Mentioned prior, but this section feels like it needs to be "you will log with json, you will use journald, you will ship your logs" type of vibe. Also the included logging refers to ziti-edge-tunnel and just doesn't feel reusable

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+## Reset
```
dovholuknf: Can we change this to "Uninstall" instead and remove the final 'install' step? (also notice it uses 'apt' not 'dnf' so removing that makes sense to me)

After this is done, I can expect all state was removed/cleaned up and nothing remains. What about my offlined PKI? Are there any other things to cleanup? Log shipping? Firewall rules, etc?

docusaurus/docs/guides/deployments/10-linux/10-controller/15-migrate.mdx
```diff
@@ -0,0 +1,48 @@
+---
+title: Migrate a Controller Installation
+sidebar_label: Migrate
+---
+
+You can generate a configuration with the `ziti create config controller` command, optionally mutating the result through command-line options or environment variables. You can also find an annotated sample config file from [the Ziti repo](https://github.com/openziti/ziti/blob/main/etc/ctrl.with.edge.yml).
```
dovholuknf: This section should just talk about "if you already have a configuration you like and you want to bring it in alignment with the new packaging, here's how you do it" vibe and not talk about `ziti create config` or any ofthat. 

If you like the answer file approach, it seems like a migration could be part of that answer file... You specify the things needed and then on startup the process 'migrates' the configuration on your behalf? (but doesn't delete the existing configuration)

That or this should just be a set of tasks, enumerated, with "for example" type commands that show them one way to accomplish the task.

docusaurus/docs/guides/deployments/20-docker/10-controller.mdx
```diff
@@ -0,0 +1,8 @@
+---
+title: Deploy the Controller with Docker
+sidebar_label: Controller
+---
+
+import DockerControllerReadme from '/_remotes/ziti-cmd/dist/docker-images/ziti-controller/README.md';
```
dovholuknf: i'm all for reusing these here but can we please remove the codeblock hints? the rendering in docusaurus is still lackluster and we don't have the time to polish it right now

qrkourier: https://github.com/openziti/ziti/pull/1987

docusaurus/docs/downloads.mdx
dovholuknf: i guess what I was trying to check/verify is that the user would `dnf install openziti` and `systemctl stop ziti-controller` or `systemctl stop ziti-router`. That's what I was trying to ask, but wasn't clear. I searched this text and now I believe I see that's the case. I do see `openziti-controller` referenced at install time, which seems reasonable.

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
+---
+title: Router Deployment
+---
+
+import CliLogin from '/docs/_cli-login.md'
+import LinuxPackageRepo from '/docs/_linux-package-repo.mdx';
+import CliLogFormats from '/docs/_cli-log-levels-and-formats.md'
+
+This article is about deploying a Ziti router as a Linux system service. [This page](/learn/introduction/03-components.md#openziti-router) provides an overview of the architecture.
```
dovholuknf: i didn't catch this before, but I'd prefer to see us use the "informal" terminology in most every case. Instead of `Ziti router` or `Ziti controller` they should be deformalized to just "router" and just "controller". I scanned through and there's a lot of places where the more formal wording is used. I'd like to turn them into the less formal tone. It struck me today when I was looking at the "Router Deployment" doc and "Ziti router" was front and center. Also, when using the formal name, I think it's best to treat it like a product name and thus it's a proper noun so both Z and R get capitalized.

dovholuknf: For the controller, you wrote "The controller introduction introduces the architectural concepts used in this guide". I really liked that wording and would perfer that vs "This page"

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+Routers provide Ziti network entry and exit points for clients and smart routing through the mesh. You must have at least one router that is advertising an edge connection to clients. If you have more than one router, you must configure transport links between them to form the fabric mesh.
```
dovholuknf: feels too reference-y to me. I like the overall vibe but I'd like to steer away from things like "you must configure transport links between them to form the fabric mesh." since this is actually entirely untrue. it's not necessary at all... :) so when it's squishy like that, i vastly prefer not to even open that door. I think this whole section could mabye even be just removed? i feel like these nuances are a whole page (or set of pages) on their own

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+## Installation
+
+The router package `openziti-router` provides a systemd service unit and optional bootstrapping. The package depends on the `openziti` package which provides the `ziti` CLI. The easiest way to install both is the cross-platform install script for RPM and Debian distributions.
```
dovholuknf: It'd be nice to show people here in the docs how to start, stop, status the package. Just naming it mihgt be sufficient? I'd think a reader just wants to know "what" they are going to `systemctl start/stop/status`. (same comment wrt controller)

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+```text
+curl -sS https://get.openziti.io/install.bash \
+| sudo bash -s openziti-router
+```
+
+<LinuxPackageRepo />
+
+## Bootstrapping
```
dovholuknf: I'd really like this to be explicitlly spelled out with steps. Here's how you add a router to your overlay network:
* create a router in the controller
* download the router's enrollment token
* create a configuration -- here's the important information you must provide
* start the router
* profit

that sort of thing. This talk of bootstrapping doesn't fit the flow to me. If there's a "streamlined" way to do things, cool but people today should be able to do all these steps manually and they should be called out explicitly here

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+You can opt-in to generating a configuration file and enrolling the router automatically. You will be prompted for the most relevant configuration values when installing interactively unless the answer is already known.
```
dovholuknf: I feel the same way about this as the controller. I feel like the service should _not_ start unless all answers are provided in some answerfile or somesuch. It's such a common problem to have the advertised addresses wrong that it should be a necessary step.

Routers are also complex to configure. One needs to understand:
* do you want this router to advertise links
* do you want this router to advertise edge

Then there's the whole: do you want this router part of a router group, what interface, what ip address....  it's just too complex imo to try to do in one big automagical bootstrapping. I'd much rather see a "here's where your configuration is, go edit it when/if you need to" and really the only thing that they need to get right is the advertised addresses because that ends up in the server certificate...

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+    1. Set `ZITI_BOOTSTRAP=true`
+1. In `/opt/openziti/etc/router/bootstrap.env` (the answer file for bootstrapping the router)
+    1. Set `ZITI_ENROLL_TOKEN` to the resulting token (JWT)
+
+Additionally, you probably want to set these:
+
+1. In `/opt/openziti/etc/router/bootstrap.env` (the answer file for bootstrapping the router)
+    1. Set `ZITI_CTRL_ADVERTISED_ADDRESS` to the FQDN of the controller (default: localhost).
+    1. Set `ZITI_ROUTER_ADVERTISED_ADDRESS` to the permanent FQDN of the router (default: localhost). This value can not be changed after enrollment.
+    1. Set `ZITI_ROUTER_MODE` (default: none) if this router's built-in tunneler will provide a proxy for services. Mode `tproxy` requires additional kernel capabilities in `/etc/systemd/system/ziti-router.service.d/override.conf` and DNS resolver configuration for the host.
```
dovholuknf: just commenting to indicate the full range of things that feels to me needs to be rearranged (based on the prior comment)

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+### Bootstrapping the Configuration
+
+The router's configuration is loaded from a YAML file ([reference](/reference/30-configuration/router.md)). The Linux system service will generate a valid configuration during the first startup, unless one already exists. You can modify the generated configuration in `/var/lib/ziti-router/config.yml`.
```
dovholuknf: same thoughts here as the controller. i think i like the idea that the configuration is generated from the answer file, which the process will then remove and putting a new answer file 'somewhere' is the indicator that you'd like to "reinstall" (or something, i'm not sold on the flow but it feels right to me at this time)... also it'd be worth calling out that when the router enrolls it will create a PKI which contains the server's advertised address in the server cert... which makes me think that we should indicate here that you basically only get ONE address for your router. Also alt server certs should be mentioned somewhere in the router page and why it's relevant and where/how and WS listeners (specifically for browzer)

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+The filename of the configuration file, relative to the router's working directory, is given as an argument, e.g., `entrypoint.bash run config.yml` in `/lib/systemd/system/ziti-router.service` as an argument to the `ExecStart` command. You may override this in `/etc/systemd/system/ziti-router.service.d/override.conf`.
```
dovholuknf: same question as before. do people _really_ need this functionality? i'd assert no, which is just one less thing to doc/test/etc?

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+Explore configuration option variables by running `ziti create config environment --help`. Any of these may be set in `/opt/openziti/etc/router/bootstrap.env` to influence the generated configuration.
```
dovholuknf: Torn on this. I think it's better to jsut omit this at this time until there's a demonstrated need for people to know it exists .. at that time we can add more /better doc. I'd prefer to see this removed

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+Disable generating a configuration by setting `ZITI_BOOTSTRAP_CONFIG=false` in `/opt/openziti/etc/router/service.env`.
```
dovholuknf: the CLI Mgmt page feels like reference info, not 'guide' info. It doesn't feel like it belongs in the guide section

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+### Bootstrapping the Enrollment
+
+You must provide an enrollment token. Obtain the token by administratively creating the router.
```
dovholuknf: This is really just "enrolling the router" imo. it's really not bootstrapping anything. There's no other way to create a router than 'administratively'. Let's remove that wording and it's a guide. I think less-formal is better here. This sentence sounds more formal/rigid to me

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+```text title="Create a router"
+ziti edge create edge-router "AcmeRouter1" \
+    --tunneler-enabled \
+    --jwt-output-file path/to/token.jwt
+```
+
+[Learn more about managing routers with the CLI](/guides/deployments/10-linux/20-router/40-cli-mgmt.md).
+
+The systemd service looks for the token in `/opt/openziti/etc/router/.token` if env var `ZITI_ENROLL_TOKEN` is empty. The file must be readable by root (not others).
```
dovholuknf: Am i not allowed to run two routers on one machine? This makes me think I can't.

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+`ZITI_ENROLL_TOKEN` may be defined in `/opt/openziti/etc/router/bootstrap.env`.
```
dovholuknf: i'm still advocating to remove the additional bootstrapping idea in general for all the same reasons already stated.

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+The router will enroll with the controller during the first startup. The one-time enrollment token is consumed during the enrollment process and a private key is generated in the router's working directory.
```
dovholuknf: It's not jsut a key. it's also two certs and a ca. I didn't see any treatment anywhere on how one backs up a router. I feel like that should be a section heading and should show up in the right nav

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+Disable bootstrapping enrollment by setting `ZITI_BOOTSTRAP_ENROLLMENT=false` in `/opt/openziti/etc/router/service.env`.
+
+## Firewall
+
+The router listens on a single configurable TCP port: `3022/tcp`. Ziti clients use ALPN to negotiate a connection to the edge (`ziti-edge`), fabric links (`ziti-link`), or health-check APIs (`h2`, `http/1.1`).
```
dovholuknf: While it's easier, and I'm 💯 in agreement to  push people to a single port, it's not necessary by any means. __IF__ someone wants to separate the link plane and edge plane ports, they most definitely can. I don't agree with the wording here. We should instead simply state that they need to ensure the firewall is open for __all__ the ports they are using. link ports, edge ports, etc. it can be many. We can tell them how to discover what ports are actually listening, maybe using `ss` maybe by 'reading the config file`...

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+You may set `ZITI_ROUTER_PORT` in `/opt/openziti/etc/router/bootstrap.env` to bootstrap with a different port.
+
+Clients "learn" the router's address and port when they poll the controller for authorized routers, so it is not necessary to re-enroll or re-create the client if the router's address or port changes. Simply change the config, bounce the router, and it will begin advertising the new address and port. This works because the system service and Docker container both auto-renew their server certificate every startup, and the certificate's DNS subject alternative name is set in the router's config.yml file. If you set `ZITI_AUTO_RENEW_CERTS=false` then it's necessary to administratively re-create the router or run at least once with the `--extend` flag.
```
dovholuknf: This information is not relevant to 'firewall' imo and is largely 'reference' too. I think this whole entire paragraph can be removed. I'd rather see an example using openssl to tell show the user HOW to verify:
a.) they can connect to the controller and on the expected ports (using openssl i'd expect)
b.) taht they should test their edge router having open ports (by again using openssl) on some OTHER machine (or on a subset of representative machines expecting to use this router)

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+## Agent
```
dovholuknf: Reference information and unnecessary in the 'install' portion. It's plausible that there's a whole troubleshooting sub page though that not only talks about HOW to use the agent (the fact that there's an 'agent' and uses IPC is reference information and not relevant to installation/management to me but maybe it's relevant insofar as permissions?)

It should cover collecting logs, acquiring a stack dump and heap dump and other such 'operational' tpye things

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+## Logging
+
+<CliLogFormats/>
+
+## Reset
+
+```text
+sudo systemctl disable --now ziti-router.service
+sudo systemctl clean --what=state ziti-router.service
+sudo apt install --reinstall openziti-router
```
dovholuknf: same comment from controller. this should tell me that it'll remove all my stuff, maybe even WARN me that it's irreversible and things like that. also the `apt installl` shouldn't be included, instead this should be titled "Uninstall" or somethign like that

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+```buttonless title="Output"
+goroutines: 38
+OS threads: 20
+GOMAXPROCS: 16
+num CPU: 16
+```
+
+## Logging
+
+<CliLogFormats/>
```
dovholuknf: All the same comments from the controller here. I don't think we can resuse this and i'd like to see treatment for log shipping added. I also want to tell them "you will use json and like it" because that's what the nfconsole does for good reasons.

docusaurus/docs/guides/deployments/10-linux/20-router/30-configuration.md
```diff
@@ -0,0 +1,228 @@
...
+id: router-configuration
+title: Router Configuration
+sidebar_label: Configuration
+hide_table_of_contents: false
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+This article expands on the [Linux and Docker router deployment](/guides/deployments/10-linux/20-router/10-deploy.mdx) article with configuration concepts and examples. [Kubernetes is covered separately](/guides/deployments/30-kubernetes//kubernetes-router.mdx).
```
dovholuknf: This whole paragraph should be removed imo and this doc moved into reference somewhere. it doesn't feel guide related at all to me and shouldn't call out linux, nor kubernetes etc.

docusaurus/docs/guides/deployments/10-linux/20-router/30-configuration.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Config Management
+
+You may create or edit the configuration file manually or run `ziti create config router` to generate one. The generator aims to accelerate developing a working configuration with the most relevant options. The output can be influenced with arguments and variables. Run `ziti create config environment` for descriptions of the available environment variables.
```
dovholuknf: In general, I'd like to avoid phrases like "you may". I'd like to favor more direct phrasing. "To create an initial configuration file, run....". This is also reference-style doc. It should be more direct and to the point. It might not even require the "To" preamble and be more direct even. "do this thing by running this command"

docusaurus/docs/guides/deployments/10-linux/20-router/30-configuration.md
```diff
@@ -0,0 +1,228 @@
...
+  xgressDialQueueLength: 1000
+  xgressDialWorkerCount: 128
+  linkDialQueueLength: 1000
+  linkDialWorkerCount: 32
+
+```
+
+</TabItem>
+
+</Tabs>
```
dovholuknf: this whole file feels like it belongs merged into the `/docs/reference/configuration/router` page and some (much?) feels redundant?

docusaurus/docs/guides/deployments/10-linux/20-router/50-migrate.mdx
```diff
@@ -0,0 +1,50 @@
...
+sudo sed -Ei 's|(ZITI_BOOTSTRAP_.*)=.*|\1=false|g' /opt/openziti/etc/router/service.env
+
+# run only the controller in the background using the quickstart state so the enrolled router can check in
+nohup ziti controller run ctrl.yaml &
+
+# start the service
+sudo systemctl enable --now ziti-router.service
+sudo systemctl status ziti-router.service
+```
+
```
dovholuknf: same comment as the controller. feels like it should be part of the other page and a set of steps with more handwavy, generaltities, "for example" commands.

docusaurus/docs/guides/deployments/20-docker/30-console.mdx
```diff
@@ -0,0 +1,67 @@
+---
+title: Console Deployment
+sidebar_label: Console
+---
+
+This article is about deploying the Ziti console as a standalone Node.js server application in a Docker container.
```
dovholuknf: I don't know why but I feel better not including the console in the whole "production install" sort of section. It feels like is should be in its own kind of section to me we already have much of this covered in the existing quickstart section. I'd feel better just removing it for now

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+## Reset
+
+```text
+sudo systemctl disable --now ziti-router.service
+sudo systemctl clean --what=state ziti-router.service
+sudo apt install --reinstall openziti-router
+```
```
dovholuknf: Thinking more about this... Router should probably also have dedicated sections for adding a the 'common' types of routers. at least "public/tunneler/edge enabled", "private/tunneler/edge enabled" and maybe (but i'm not sold on this) "fabric only"?

docusaurus/docs/guides/index.md
dovholuknf: conversation marked as resolved, but the change still shows?

docusaurus/docs/guides/index.md
```diff
@@ -3,13 +3,15 @@ title: Overview
...
 
 ## Add or Request New Guides
 
 There's lots of room in here for new guides! Want guidance before you submit? Want someone else to submit? No problem. Follow the link to create a GitHub issue, label it "how-to", and tell us what you are working on. 
 
 * [Request a new guide or find guide requests](https://github.com/openziti/ziti-doc/issues?q=is%3Aissue+is%3Aopen+label%3Ahow-to+)
 * [Where in GitHub to add or edit a guide](https://github.com/openziti/ziti-doc/tree/main/docusaurus/docs/guides)
+
+## What is a Guide?
```
dovholuknf: I'd like to see this where it used to be and the other text removed or incorporated into this

docusaurus/docs/reference/backup/10-controller.mdx
```diff
@@ -0,0 +1,87 @@
+---
+title: Controller Backup
+sidebar_label: Controller
+---
+
+## Backup
+
+The best way to ensure continuity is to run multiple controllers. You can recover from the catastrophic loss of all
```
dovholuknf: This can't merge yet

docusaurus/docs/reference/backup/10-controller.mdx
```diff
@@ -0,0 +1,87 @@
...
+controllers in the cluster by creating a new single controller which can become the inaugural leader of a new cluster.
+
+The three stateful aspects of the controller are the configuration YAML, PKI, and the
+database. Ensure you back up all the files mentioned in the controller's configuration YAML file (keys, certs, root CA
+bundle, and the database), and the config YAML file itself.
+
+If you're using the Linux service you may accomplish this by triggering the database snapshot operation and backing up
+the `/var/lib/ziti-controller` directory.
+
+### Backing up the Configuration YAML
```
dovholuknf: This all feels like "guide related" and not reference. A reference doc shouldn't talk about how to back something up, it should just tell you "this file is important and it's located here and it has this purpose"... This content feels more like it belongs in the guide section and is probably bespoke per install (linux/docker/kubernetes)

docusaurus/docs/reference/backup/20-router.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+The router's identity files (keys and certs) uniquely identify the router to the controller. To recover a unique router
+identity you must back up the files mentioned in the following configuration properties:
+
+- `identity` may appear in several places and always configures keys and certs for the router
+
+### Configuration YAML
+
+The router's configuration YAML file is specified as a positional parameter when the router is run. You can edit the
+file to change the filesystem path for any files that are restored to a new location.
```
dovholuknf: some comment here as backup->controller. all feels like it fits in each 'deployment' section and bespoke to each.

docusaurus/docs/reference/tunnelers/60-linux/30-manual-installation.mdx
```diff
@@ -0,0 +1,35 @@
...
+sidebar_label: Manual
+---
+
+## Manual Installation
+
+[The latest binary release](https://github.com/openziti/ziti-tunnel-sdk-c/releases/latest/) of `ziti-edge-tunnel` is
+distributed as an executable for amd64, arm, and arm64 architectures. To upgrade the tunneller perform the installation
+procedure again.
+
+Install the `wget` and `unzip` commands to use this example.
```
dovholuknf: I actually feel like the 'manual' installation should be a "go download it from github, here's the link" and leave it at that. You don't need to add what arch's it's distributed for. if they change or get added, the information is out of date and people will figure it out. vagueness is better by saying "most popular linux variants" or something handwavey like that. 

i kinda think the whole page should be more like a footnote

docusaurus/docs/reference/config-types/_category_.yml
```diff
@@ -1 +1 @@
-label: Builtin Config Types
+label: Config Types
```
dovholuknf: i'd prefer to leave "Built-in"

docusaurus/docs/reference/tunnelers/80-kubernetes/index.mdx
```diff
@@ -1,5 +1,6 @@
 ---
-title: Workload Tunneling
+title: Tunneling Kubernetes Workloads
```
dovholuknf: The previous title works better for me. It's kinda redundant to see Kubernetes in the title when i just clicked the Kubernetes sidebar thingy. Also it makes the title wrap even when the windows is modestly sized. I'd like to revert this

docusaurus/docs/guides/05-troubleshooting/_category_.yml
qrkourier: this sidebar category was moved, not deleted

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: Maybe firewall rules. All state is destroyed by the clean command.

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: The title should be descriptive without being too long. The title is separate from the breadcrumb/sidebar hierarchy, i.e., the title stands alone. Conversely, the `id` (slug) is part of the breadcrumb (in the context of sidebar labels) and shouldn't be redundant, so the `id: deploy` should have been just `index`.

That is `guides/deployments/linux/controller/deploy` is redundant, but it didn't seem worth changing.

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: That's done by running install.bash. If they opt for the manual approach (not install.bash) then it's safe to assume they know about the install command.

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: Raised an issue for supporting IPv4 addresses: https://github.com/openziti/ziti/issues/1993 and documenting that option in https://github.com/openziti/ziti-doc/issues/833.

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: issue for changing controller leaf cert renewal from startup to interval: https://github.com/openziti/ziti/issues/1994

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: For the Linux package described here, generating a configuration is opt-in. The steps to enable it are in the section about generating a config YAML.

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: ~~Are you referring to the firewall section?~~ NM. It's about the agent section.

docusaurus/docs/_cli-log-levels-and-formats.md
qrkourier: Deleted this file and duplicated and tailored the content in controller and router deploy guides where it was being included as a partial.

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: There's no obvious home for this kind of info in the `/reference` section yet, so I've moved it to `/learn` for now.

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: moved to "enrollment" article

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: moved to troubleshooting

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: raised an issue to redress how db bootstrapping works in https://github.com/openziti/ziti/issues/2000

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: My goal was to answer the predictable questions "Where is the config.yml file?" and "What determines the filename 'config.yml'?"

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: Let me know what you think about the new presentation.

docusaurus/docs/reference/backup/10-controller.mdx
qrkourier: We decided to move it from guides to reference.

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: Added link to the article where securing the root CA is documented as part of controller backup.

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: If I understand your preference to remove information about opt-in to bootstrapping, then I'm unsure how to offer both alternatives. This section is about providing or generating a configuration (your preferred umbrella term for bootstrapping). If they provide a configuration, then the default service configuration works (no opt-in to bootstrapping). If they are not providing any one of the three pillars of the configuration, then they must opt-in to bootstrapping by setting the var in service.env.

docusaurus/docs/reference/backup/10-controller.mdx
qrkourier: I'll remove references to multiple controllers.

docusaurus/docs/guides/deployments/10-linux/10-controller/15-migrate.mdx
qrkourier: I like the idea of fragmenting this BASH script into a more instructive set of steps.

Migrating a complete configuration and the bootstrapping answer file are mutually-exclusive. Migrating a partial configuration (e.g., migrating only the db, but generating the rest) would involve answering the bootstrapping questions pertinent to the parts of the configuration that are to be generated (e.g., the PKI and config.yml), and any answers pertaining to the migrated db (i.e., `ZITI_PWD`) would be ignored during bootstrapping.

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: I'm hearing two suggestions: require an answer file when not providing a configuration, and remove the answer file (including the admin password if it was provided there) when the configuration has been generated.

The answer file (`bootstrap.env`) is created without answers (null values) when the controller is installed. If it is missing, then systemd will fail to start the service because this answer file is one of the required env files.

If `bootstrap.env` does not contain the required answers (e.g., `ZITI_CTRL_ADVERTISED_ADDRESS`), then service will not start because bootstrapping will return an error.

The answer file can only be removed by refactoring the service unit and entrypoint script because it is a required env file for the service unit and life cycle scriptlets rely on the service.env and bootstrap.env answer/env files to determine whether or not a particular question is already answered. For example, if bootstrapping is enabled and a question is not answered then the package manager upgrade scriptlet may then prompt for the answer or take background action to correct an ambient kernel capability. So, the answer files are parcel to the package life cycle.

I raised an issue to redress how db bootstrapping works in https://github.com/openziti/ziti/issues/2000 in response to the suggestion that the Linux package should proactively delete the ziti password from disk after the database is initialized.

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: issue for supporting alt server certs: https://github.com/openziti/ziti/issues/2001 (is linked to documentation counterpart issue)

docusaurus/docs/reference/tunnelers/60-linux/50-linux-tunnel-troubleshooting.mdx
```diff
@@ -16,13 +16,11 @@ The tunneler obeys the value of `logLevel` in `/var/lib/ziti/config.json`. The i
 Create a log file from the current systemd service invocation to share with collaborators.
 
 ```bash
```
dovholuknf: bash --> text pls

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
qrkourier: Good call. Raised this issue to track adding support for private routers with a linked doc issue for adding those guides: https://github.com/openziti/ziti/issues/2003

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
qrkourier: I do see an opportunity to split this info between the part that's pertinent to the type of deployment (Linux, using `nsenter`) and the part that's generic to all types of deployments (the agent commands you can run for ops tasks, not just troubleshooting).

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
qrkourier: I found the router will start without the ctrl address and raised this issue to fix it: https://github.com/openziti/ziti/issues/2005

qrkourier: I agree. There's no benefit to wrapping every possible config property with this package's bootstrapper or something like `ziti_router_auto_enroll`. Rather, we should document how to configure the router's YAML file for each scenario, whether that config was written from scratch or generated by the bootstrapper as a starting point.

I'll explore in that linked GH issue which answers should be required.

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
qrkourier: https://github.com/openziti/ziti-doc/pull/848

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
qrkourier: https://github.com/openziti/ziti-doc/pull/854

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
qrkourier: I think I see what you mean. With multiple routers, it is possible for them to provide only the edge without any fabric. Services will still work as long as there's one common, online router between the dial and bind identities. Smart routing via the mesh fabric is, technically, optional.

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
qrkourier: It is necessary for this Linux router because that's how it works currently. Here's a feature request for discrete ziti-edge and ziti-link when using the router's bootstrapper: https://github.com/openziti/ziti/issues/2014

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
qrkourier: https://github.com/openziti/ziti-doc/pull/855

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
qrkourier: https://github.com/openziti/ziti-doc/pull/855

qrkourier: https://github.com/openziti/ziti-doc/pull/858

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
qrkourier: https://github.com/openziti/ziti-doc/pull/858

PR #823, opened 2024-04-16
add category titles, descriptions, distinguishing slugs as needed

qrkourier:
My goal is to make the generated index pages more helpful and to distinguish their category slugs if they have generic default slugs.


dovholuknf:
I'd really rather not change the slugs, please. it's not helping anything in reality and it's just going to lead to broken links. let's not?

PR #847, opened 2024-05-01
install in two steps to allow for interactive prompts

qrkourier:


dovholuknf (dismissed):
i'd prefer to see this be straightforward with a clear set of instructions, and not offer a choice. "do it this way" -- not "if you do it this way"

dovholuknf (approved):

PR #848, opened 2024-05-01
move agent info to troubleshooting

qrkourier:


dovholuknf (approved):
makes perfect sense being in troubleshooting like this. thx

PR #858, opened 2024-05-03
help the linux router user verify firewall ports

qrkourier:


docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -82,7 +82,42 @@ Disable bootstrapping enrollment by setting `ZITI_BOOTSTRAP_ENROLLMENT=false` in
 
 ## Firewall
 
-The router's generated configuration always listens on a single, configurable TCP port (default: `3022`). You may set `ZITI_ROUTER_PORT` in `/opt/openziti/etc/router/bootstrap.env` to bootstrap with a different port.
+The router's generated configuration always uses a single TCP port (default: `3022`) defined by `ZITI_ROUTER_PORT` in `/opt/openziti/etc/router/bootstrap.env`.
```
dovholuknf: would you remove the word 'always'? it's an extra word that doesn't add any additional clarity.

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -82,7 +82,42 @@ Disable bootstrapping enrollment by setting `ZITI_BOOTSTRAP_ENROLLMENT=false` in
...
+Substitute the router's advertised address and port for `localhost:3022` as defined in `/var/lib/ziti-router/config.yml`.
+
+```text
+EXTERNAL_URI=localhost:3022
+```
+
+Verify the router is reachable and presents a server certificate from the controller's intermediate edge signer CA for both ALPN identifiers: `ziti-edge` and `ziti-link`.
+
+```text
+for ALPN in ziti-{edge,link}; do
```
dovholuknf: it's better to be straightforward and not terse/elegant by using a for loop. please split this into two separate commands which are easy to explain to the reader and easy to follow.

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -82,7 +82,42 @@ Disable bootstrapping enrollment by setting `ZITI_BOOTSTRAP_ENROLLMENT=false` in
 
 ## Firewall
 
-The router's generated configuration always listens on a single, configurable TCP port (default: `3022`). You may set `ZITI_ROUTER_PORT` in `/opt/openziti/etc/router/bootstrap.env` to bootstrap with a different port.
+The router's generated configuration always uses a single TCP port (default: `3022`) defined by `ZITI_ROUTER_PORT` in `/opt/openziti/etc/router/bootstrap.env`.
+
+Confirm the router is running and listening on the expected port.
+
+```bash
+systemctl show -p MainPID --value ziti-router.service \
```
dovholuknf: isn't there any more straightforward way to do this? You've already declared the port up above. it would be simple to simply `netstat | grep {that port}` no? why all this complexity when a simple netstat would suffice? is there a reason?

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -82,7 +82,42 @@ Disable bootstrapping enrollment by setting `ZITI_BOOTSTRAP_ENROLLMENT=false` in
...
+| xargs -rIPID sudo lsof -Pnp PID \
+|& grep LISTEN
+```
+
+```buttonless title="Output"
+ziti    3574455 ziti-router    7u     IPv6           41543151       0t0      TCP *:3022 (LISTEN)
+```
+
+Substitute the router's advertised address and port for `localhost:3022` as defined in `/var/lib/ziti-router/config.yml`.
```
dovholuknf: "3022" should be replaced with whatever their port is  and should be referenced here to remind them that it's not always 3022

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
qrkourier: I can't think of a more straightforward to answer the question "Is the PID of the router service listening on any ports?" Checking for something listening on some port, even if that something is named "ziti" doesn't tell me if that ziti is the router service, or some other router container or native process.

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
```diff
@@ -82,7 +82,34 @@ Disable bootstrapping enrollment by setting `ZITI_BOOTSTRAP_ENROLLMENT=false` in
...
+```
+
+```buttonless title="Output"
+ziti    3574455 ziti-router    7u     IPv6           41543151       0t0      TCP *:3022 (LISTEN)
+```
+
+Verify the router is reachable and presents a server certificate from the controller's intermediate edge signer CA and the intermediate CA cert. Substitute the router's advertised address and port for `localhost:3022` with the advertised address in `/var/lib/ziti-router/config.yml`.
+
+```text
+openssl s_client -connect localhost:3022 -alpn ziti-ctrl -showcerts <>/dev/null \
```
qrkourier: Use `ziti-edge`, not ctrl

docusaurus/docs/guides/deployments/10-linux/20-router/10-deploy.mdx
dovholuknf: It's just looking at this bash is a mind-twister for people who aren't well-versed with bash. imo, this is too complex for the sake of copy/paste simplicity. We don't need the doc to be bullet proof, it needs to be understandable. using advanced (xargs and lsof, imo are 'advanced') bash to combine three commands into one copy/paste won't educate them and doesn't provide them the necessary context they really need and I'm certain many users will stumble over this.

I aslo assert that this can be simplified into just "make sure the service is running" by using systemctl --status. If it's running, it's listening on the configured port. Then, netstat or ss are going to report correctly that it's actually listening.

Really though, i think __the__ best thing to do is an openssl s_client -connect command. That also (within reason, imo) verifies all the firewalls between end-users and that router will be open to allow traffic through. That command is already shown below and accomplishes this same goal (only it used localhost not the FQDN)

qrkourier: I explained the one-liner and added a simplified, though not authoritative listening port check: `sudo ss -tlnp | grep ziti`.

dovholuknf (approved):

PR #864, opened 2024-05-13
update docker console guide

qrkourier:


docusaurus/docs/guides/deployments/20-docker/30-console.mdx
```diff
@@ -1,67 +1,33 @@
...
-    ]
-}
-```
+1. Log in to the console.
 
-Link to [settings template in the repository](https://github.com/openziti/ziti-console/blob/main/projects/ziti-console-lib/src/lib/assets/data/settings.json).
+    The only available authentication method for the console SPA is username and password.
 
-The Docker image's working directory is `/usr/src/app`, so the `settings.json` file may be mounted and referenced with the environment variable.
+1. Optionally, configure a console certificate.
```
dovholuknf: I think here's it's appropriate to explain to them what this means. Something like:

> The console application will now be served by the OpenZiti Controller. The controller can be configured with multiple server certificates allowing for your controller to have one url, while your console could have another. For more information about usig alternative server certs with the controller, see: {link here}

something like that

dovholuknf (approved):

PR #883, opened 2024-05-22
incorporate feedback on logging section

qrkourier:


dovholuknf (approved):
still seems like it could be presented better. something about it just doesn't look clean to me. i'd try removing the tabs maybe

PR #884, opened 2024-05-22
doc linux upgrades

qrkourier:


dovholuknf:
i dislike this pr. we don't need to tell people how to use apt/yum/etc. i won't approve these changes

PR #903, opened 2024-07-05
align linux install with package refactor

qrkourier:


docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -18,85 +18,43 @@ We'll cover the following topics:
 
 ## Installation
```
dovholuknf: With fresh eyes, I feel like this should be "package installation". This whole guide to me is "openziti controller installation" and separating the concept feels useful

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -18,85 +18,43 @@ We'll cover the following topics:
...
-wget https://get.openziti.io/install.bash
+curl -sS https://get.openziti.io/install.bash | sudo bash -s openziti-controller
 ```
 
-Install interactively so that you will have an opportunity to answer questions about generating a configuration.
-
-```text
-sudo bash ./install.bash openziti-controller
-```
+Or, manually configure the package repo and install `openziti-controller`.
```
dovholuknf: somehow, visually, this is not as clear as it could be. did you try using tabs here? it might just be our current styling, it might be presentation, but right now it doesn't feel like an "or" .  Maybe three tabs with the script install being the default? 
| script install | Debian Family | RedHat Family | 
maybe?

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -18,85 +18,43 @@ We'll cover the following topics:
...
 
 <LinuxPackageRepo />
 
+The `openziti` package provides the `ziti` CLI and is installed as a dependency.
+
 ## Configuration
 
-You must provide or generate a configuration. Configuration consists of a PKI, a config YAML file, and a database.
+You must generate, migrate, or craft a configuration. Configuration consists of a PKI, a config YAML file, and a database.
```
dovholuknf, line 41: in practicality, migrating or creating (the word 'craft' feels odd for some reason) are the same thing to me. Can we just eliminate this and go right "generate a configuration" and leave some kind of footnote/link to how to migrate/edit the configuration after it's generated? i don't think it deserves a 'heading' in this case

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
```diff
@@ -18,85 +18,43 @@ We'll cover the following topics:
...
-#### Generating a Configuration File
-
-The controller's configuration is loaded from a file ([reference](/reference/30-configuration/controller.md)). The Linux system service will generate a valid configuration during the first startup unless one already exists.
-
-The filename of the configuration file, relative to the Linux service's working directory (default: **/var/lib/ziti-controller/config.yml**). You may override this in **/etc/systemd/system/ziti-controller.service.d/override.conf**.
+This is the recommended approach if you are installing a new controller.
 
-Disable bootstrapping a configuration by setting `ZITI_BOOTSTRAP_CONFIG=false` in **/opt/openziti/etc/controller/service.env**.
+1. Set the required values in the answer file **/opt/openziti/etc/controller/bootstrap.env**.
+    1. `ZITI_CTRL_ADVERTISED_ADDRESS` - control plane permanent DNS name
```
dovholuknf: so after install, the user would need to edit this file, that's the idea, right? if that's the case, I'd like this to say "edit the answer file and set the required values".  I'd also like to eliminate the ZITI_CTRL_BIND_ADDRESS from this list since i predict < 10% of people will actually care (proabably much less).

Also the ZITI_CTRL_ADVERTISED_ADDRESS can it be: "the address all participants in the overlay will use to reach the controller"? It cannot be mandated that it's a DNS name

Finally, the easy path should be to just run the script and answer the answers when prompted. i don't see that mentioned in here yet, right? Interactive is by far the easiest for people imo

qrkourier: It must be a DNS name until https://github.com/openziti/ziti/issues/1993

docusaurus/docs/guides/deployments/10-linux/10-controller/10-deploy.mdx
qrkourier: I'd forgotten to finish adding the sections for these three options which are the same three config options for controller and router.

qrkourier (commented):
Thanks for the input. I feel like this is ready.

dovholuknf (approved):

PR #916, opened 2024-07-23
browZer IdP docs

rentallect:


docusaurus/docs/guides/Identity_Providers/_category_.yml
```diff
@@ -0,0 +1,5 @@
+label: BrowZer Identity Providers
```
qrkourier: I like the shorter "BrowZer" for the sidebar category label, especially on mobile.

docusaurus/docs/guides/Identity_Providers/index.md
```diff
@@ -0,0 +1,118 @@
+---
+title: Identity Providers for BrowZer
+sidebar_label: Identity Providers for BrowZer
```
qrkourier: This `sidebar_label` is currently being ignored because docusaurus/docs/guides/Identity_Providers/_category_.yml has `link.id: guides/Identity_Providers/index`, so that category's `label` property has precedence.

docusaurus/docs/guides/Identity_Providers/_category_.yml
rentallect: how 'bout we go with `BrowZer IdPs`?  It will be shorter while still indicating the main topic, which is all about IdPs for BrowZer.

dovholuknf (commented):
is there any way to have the 'unimplemented' instructions have more meat behind them even if it is just a generic "you need to do this / that" sort of instructions? i would think that we can very generically indicate what is necessary to support an IdP, right?

docusaurus/docs/guides/Identity_Providers/autheila/index.md
```diff
@@ -0,0 +1,18 @@
...
+  <meta
+    name="description"
+    content="How to configure Authelia for OpenZiti BrowZer."
+  />
+</head>
+
+<img src="/icons/logo-authelia.svg" alt="How to configure Authelia for OpenZiti BrowZer" width="20%"/>
+
+### Express Your Interest in this Identity Provider
+If you want to use this identity provider with your BrowZer network, please express your interest over in our [Discourse forum](https://openziti.discourse.group/).
```
dovholuknf: all of these sections might be better served from a docusaurus "partial" where you can include the exact same content in a page as opposed to duplicating it. Please consider extracting this into a partial. An example is `./_linux-package-repo.mdx` and 
```
guides/deployments/10-linux/10-controller/10-deploy.mdx
guides/deployments/10-linux/20-router/10-deploy.mdx
```

rentallect: Done.

docusaurus/docs/guides/Identity_Providers/_category_.yml
dovholuknf: it can't be "just" browzer for sure. BrowZer IdPs is probably fine. i don't agree that on mobile it makes a difference. for me it would be fine spelled out

dovholuknf (approved):

PR #923, opened 2024-07-25
un-squiggle the landing page

qrkourier:


docusaurus/static/landing.html
```diff
@@ -491,7 +535,7 @@
 			<div class="elementor-widget-wrap elementor-element-populated">
 								<div class="elementor-element elementor-element-b0a6294 elementor-widget__width-auto pink-to-purple elementor-widget elementor-widget-heading" data-id="b0a6294" data-element_type="widget" data-widget_type="heading.default">
 				<div class="elementor-widget-container">
-			<h2 class="elementor-heading-title elementor-size-default">2k+</h2>		</div>
+			<h2 class="elementor-heading-title elementor-size-default">1.7k+</h2>		</div>
```
dovholuknf: need to be careful when touching this page. might as well update this to 2.3k+

qrkourier: It may have been updated after it was exported. I'll change it in Elementor.

dovholuknf: You can but i hope that this page just stops being produced from there eventually.

dovholuknf (approved):

PR #946, opened 2024-09-13
docs tweaks

qrkourier:
These tweaks:

- reduce confusion about a router that will not start because it was created with the tunneler disabled
- recommend creating an admin identity for cert auth with new CLI login `--file` option
- recommend using cert auth with prod console


qrkourier:
~~This shouldn't merge until https://github.com/openziti/ziti/pull/2335 which is waiting for the next ziti-console release.~~

EDIT: now unblocked

qrkourier:
~~Blocked waiting for v1.1.13 introducing the login --file options.~~

EDIT: now unblocked because the changes documenting cert auth were pruned

dovholuknf (approved):

PR #978, opened 2024-11-17
use 'api://<clientId>' for 'PKCE scope' with Entra

rentallect:


dovholuknf (dismissed):

dovholuknf:
can you change from blur to a black bar or something? blur is not destructive. if you wish it to be secret, it really needs an opaque box

dovholuknf (approved):

PR #986, opened 2024-11-26
fix quickstart doc

dovholuknf:
rearrange env vars to make sense and align better. also update zac install page


corbinday:
thanks!

smilindave26 (approved):

PR #1005, opened 2025-01-13
fix: example for portChecks not showing a list of checks for portChecks

nenkoru:
Noticed a missing list reference for portChecks


dovholuknf (approved):
thanks

PR #1025, opened 2025-01-29
Add doc for configuring IdPs with ext-jwt-signers

### LINKED ISSUE #1022: Add a guide for using IdPs
configuring IdPs continues to be a challenge. We have some good starter doc for browzer but we need to bolster the doc and make it more generic.

Make doc that illustrates how to use OpenZiti for external auth from the following authorization servers:

* Google
* Entra
* Authentik
* Auth0
* Okta
* Keycloak
* Zitadel
* Incognito

After these look at the other ones BrowZer does and bring those in

dovholuknf:
closes #1022


docusaurus/docs/learn/formatting-demo.md
```diff
@@ -0,0 +1,66 @@
+---
+draft: 'true'
+---
+
+# Code Block Formatting Demo
+
+This here is a "plain" code block
+```
```
qrkourier, line 12: it will help to distinguish fence start and end if we model and generally require a non-empty value for lang, e.g., ` ```text`

qrkourier (dismissed):

smilindave26 (approved):

PR #1031, opened 2025-02-02
Add Controller HA reference material. Fixes #929

### LINKED ISSUE #929: Ensure ziti-doc is up to date for HA
Move/copy some of the ha doc from ziti to ziti-doc
  * Make sure everything is up-to-date
  * Need migration guide
  * Need overview
  * Best practices?
  
Make sure controller/router config doc is updated

Completed
* Events Configuration

plorenz:


docusaurus/docs/reference/ha/overview.md
```diff
@@ -0,0 +1,55 @@
+---
+sidebar_label: Overview
+sidebar_position: 05
+---
+
+# Controller HA 
+
+## Overview
+
+Ziti controllers can be run in a cluster for high availablity and performance scaling.
```
dovholuknf: this needs an actual "overview" of what it even means to be HA. I like the warning, it just feels like extra words are necessary describing what this means. not so much 'why', just 'what'

docusaurus/docs/reference/ha/overview.md
```diff
@@ -0,0 +1,55 @@
...
+
+:::warning
+
+**NOTE: Controller HA is still in Beta** 
+
+It's quite functional now, but we are continuing to test and refine before we mark it GA.
+:::
+
+
+### For SDK Clients/Tunnelers
```
dovholuknf: Turns out a lot of the 'what' is here under this heading. I'd say just remove this heading and it reads a bit better. the lead sentence needs a bit more gusto though imo. Just a bit of prose before jumping into benefits...

Also benefits/limitations are reasonable sub-headings here imo

docusaurus/docs/reference/ha/overview.md
```diff
@@ -0,0 +1,55 @@
...
+    1. Service lookups
+    1. Session creation
+1. Horizontal scaling of circuit creation
+
+This means that for everything that SDK clients and tunnelers depend on, controllers
+can be scaled up and placed strategically to meet user demand. 
+
+The following limitations currently apply:
+
+1. Circuits are owned by a controller. If the controller goes down, the circuit
```
dovholuknf: are there issues to track when these limitations will be eliminated? I'd like to see some text about how these are going to be addressed in the future. ideally with a link they reader could use to track if they wanted. also, removing the doc about limitations should be tracked too. feels like the kinda thing that can be easily forgotten

docusaurus/docs/reference/ha/overview.md
```diff
@@ -0,0 +1,55 @@
...
+
+1. Circuits are owned by a controller. If the controller goes down, the circuit 
+   will remain up, but can't be re-routed for performance or if a router goes down.
+2. For a controller to route circuits on a router, that router must be connected
+   to that controller. This means that routers should generally be connected to
+   all controllers.
+
+### For Management Operations
+
+The HA controller cluster uses a distributed journal keep the data model synchronized across controllers.
```
dovholuknf: "uses a distributed journal", i think it'd be good to have a link here to either raft itself or wiki about what a distributed journal is. somewhere in this overview though, we should mention that raft is used and is the distributed journal implemenation. just a mention, not in-depth.

docusaurus/docs/reference/ha/overview.md
```diff
@@ -0,0 +1,55 @@
...
+2. For a controller to route circuits on a router, that router must be connected
+   to that controller. This means that routers should generally be connected to
+   all controllers.
+
+### For Management Operations
+
+The HA controller cluster uses a distributed journal keep the data model synchronized across controllers. 
+This has the following ramifications:
+
+1. Read operations will work on any controller that is up. If the controller is
```
dovholuknf: "that is up" seems like redundant information. isn't that implied? :)

docusaurus/docs/reference/ha/overview.md
```diff
@@ -0,0 +1,55 @@
...
+   to that controller. This means that routers should generally be connected to
+   all controllers.
+
+### For Management Operations
+
+The HA controller cluster uses a distributed journal keep the data model synchronized across controllers. 
+This has the following ramifications:
+
+1. Read operations will work on any controller that is up. If the controller is 
+   disconnected from the cluster, the reads may return data that is out of date.
```
dovholuknf: > "may return data that is out of date."

does (or will) the ziti cli be able to detect this? if not, I'd like to have an issue that tracks this somehow. mabye a header in all API payloads makes sense? Something the user could use to know if their reads are possibly stale

docusaurus/docs/reference/ha/overview.md
```diff
@@ -0,0 +1,55 @@
...
+2. Update operations require that the cluster has a leader and that a quorum of nodes
+   is available. A quorum for a cluster of size N is (N/2)+1. This means that a 3 node
```
dovholuknf: does this mean it's impossible to run "HA" with a single (1) node? I realize one node isn't 'HA', but what I want to know here is if, as user, i can start with a single node or not. and if i start with a single node what ramifications will this have (if any). I am reading the doc and commenting as I go so if this is addressed elsewhere, a link to that might be useful here and something to the effect that yes you can run HA with one node makes sense to see here, imo...

dovholuknf: Also this might be just better as a link to some external doc. Regardless if you keep the text or not, a link to the doc is a good idea here

docusaurus/docs/reference/ha/overview.md
```diff
@@ -0,0 +1,55 @@
...
+The HA controller cluster uses a distributed journal keep the data model synchronized across controllers. 
+This has the following ramifications:
+
+1. Read operations will work on any controller that is up. If the controller is 
+   disconnected from the cluster, the reads may return data that is out of date.
+2. Update operations require that the cluster has a leader and that a quorum of nodes
+   is available. A quorum for a cluster of size N is (N/2)+1. This means that a 3 node 
+   cluster can operate with 2 nodes and a 5 node cluster can operate with 3 nodes, and 
+   so on. 
+3. Updates can be initiated on any controller, they will be forwarded to the leader to
```
dovholuknf: people will ask about what this means. will the ziti cli wait for quorum? do we have to explain eventual consistency here? is there any way to wait/check that these values have been successfully processed by other nodes? again, if explained in more detail elsewhere a link with 'more information' should be here

docusaurus/docs/reference/ha/overview.md
```diff
@@ -0,0 +1,55 @@
...
+   be applied.
+4. The cluster may have non-voting members.
```
dovholuknf: This factual comment leaves the reader wondering what this actually means.  Would you add a bit more detail on this and why a non-voting member might be useful?

docusaurus/docs/reference/ha/overview.md
```diff
@@ -0,0 +1,55 @@
...
+   disconnected from the cluster, the reads may return data that is out of date.
+2. Update operations require that the cluster has a leader and that a quorum of nodes
+   is available. A quorum for a cluster of size N is (N/2)+1. This means that a 3 node 
+   cluster can operate with 2 nodes and a 5 node cluster can operate with 3 nodes, and 
+   so on. 
+3. Updates can be initiated on any controller, they will be forwarded to the leader to
+   be applied.
+4. The cluster may have non-voting members. 
+
+See [topology](./topology.md) and [the data model](./data-model.md) for more information.
```
dovholuknf: ah. here are some links. can these be adjacent to where the information is relevant? anything to get the links closer to where the idea/concept is useful is helpful for people to find the right links

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+
+# Bootstrapping A Cluster 
+
+To bring up a controller cluster, one starts with a single node. 
+
+## Controller Configuration
+
+### Certificates
+
+Each controller requires appropriate certificates. The certificates for clustered controllers
```
dovholuknf: As this appears to be a pre-requisite, it feels like it should be part of the bootstrapping section. Maybe it makes sense to turn this into a group and have the steps of bootstrapping called out as individual left-nav sections.

Bootstrapping
* creating certs
* update config file
* run controller
* initialize controller
* add nodes
* removing nodes

Also this all seems like 'operations' to me. Looking at the 'Operations' page, other than restoring backup and open ports, it seems more like it might be a 'monitoring' type of page?

Feels like grouping some of this would help.  Maybe "Configuring" vs "Running|Monitoring"?

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+
+The controller requires a `raft` section.
+
+```yaml
+raft:
+  dataDir: /path/to/data/dir
+```
+The `dataDir` will be used to store the following:
+
+* `ctrl-ha.db` - the ziti model bbolt database
```
dovholuknf: ctrl-ha.db, raft.db, shapshots are all mentioned here but the yaml section above contains none of these values?

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+* `raft.db` - the raft bbolt database
+* `snapshots/` - a directory to store raft snapshots
+
+Controller use the control channel listener to communicate with each other. Unlike
```
dovholuknf: > Controller use the

--> The controller uses the 

? maybe "Controllers use the"?

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+routers, they need to know how to reach each other, so an advertise address must
+be configured.
```
dovholuknf: > so an advertise address must be configured.

On an implementation pov, I don't understand this because it feels like a chicken-egg situation. As a reader, you just told me they need to know how to reach each other but then tell me to update the controller to advertise a particular address. It's not clear to me how this helps a joiner... Didn't I have to tell the joiner to join an existing cluster? This seems odd?

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+
+```yaml
+ctrl:
+  listener: tls:0.0.0.0:6262
+  options:
+    advertiseAddress: tls:192.168.1.100:6262
+```
+ 
+Finally, for sessions to work across controllers, JWTs are used. To enable these
```
dovholuknf: > for sessions to work across controllers, JWTs are used.

Is this relevant information at this point? It feels like this only makes me ask more questions and makes me want more information. I feel like this should be a "do this" section: "enable the OIDC endpoint" and then you don't have to answer the question about 'why' -- well... we'll see i guess? :)

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+      - binding: health-checks
+      - binding: fabric
+      - binding: edge-management
+      - binding: edge-client
+      - binding: edge-oidc
+```
+
+## Initializing the Controller
+
+Once properly configured, the controller can be started.
```
dovholuknf: > Once properly configured

Makes me think -- this will be easy to misconfigure. If there are misconfigurations, a troubleshooting guide would be good to see. I suspect there's not one yet so maybe just say "start the controller"?

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+
+```shell
+ziti controller run ctrl1.yml
+```
+
+Once the controller is up and running, it will see that it is not yet initialized, and will pause
```
dovholuknf: > Once the controller is up and running, it will see that it is not yet initialized

Should this say "the first time the controller starts, it will see that it's not yet initialized"?

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+      maxTLSVersion: TLS1.3
+    apis:
+      - binding: health-checks
+      - binding: fabric
+      - binding: edge-management
+      - binding: edge-client
+      - binding: edge-oidc
+```
+
+## Initializing the Controller
```
dovholuknf: > ## Initializing the Controller

--> ## Initializing the First Controller

?

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+
+```
+[   3.323] WARNING ziti/controller/server.(*Controller).checkEdgeInitialized: the 
+Ziti Edge has not been initialized, no default admin exists.  Add this node to a 
+cluster using 'ziti agent cluster add tls:localhost:6262' against an existing 
+cluster member, or if this is the bootstrap node, run 'ziti agent controller init' 
+to configure the default admin and bootstrap the cluster
+```
+
+As this is the first node in the cluster, we can't add any nodes to it yet. Instead, run:
```
dovholuknf: This is the first time i've see a "we" or a "you". I try to avoid these words in 'reference' doc.

> As this is the first node in the cluster, we can't add any nodes to it yet.

--> `As this is the first node in the cluster, nodes cannot be added to it before it is initialized. Run the following command to initialze the first controller`

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+
+```
+ziti agent controller init <admin username> <admin password> <admin identity name>
+```
+
+This initializes an admin user that can be used to manage the network.
+ 
+## Managing the Cluster
```
dovholuknf: > ## Managing the Cluster

This title doesn't feel like 'bootstrapping' to me. Maybe this is a separate page?

dovholuknf: Maybe the whole "Bootstrapping" should be "Managing the Cluster" and a page added for "first cluster"?

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+
+There are four commands which can be used to manage the cluster.
```
dovholuknf: `ziti agent` implies to me you MUST be on the same machine as 'the controller' because it's using an IPC socket. Is that true? Is cluster management exclusively by "a machine running the controller"? That's something i haven't seen before when reading the doc and should be made clear if that's the case?

docusaurus/docs/reference/ha/overview.md
```diff
@@ -0,0 +1,55 @@
...
+
+A controller cluster offers the following advantages:
+
+1. Horizontal scaling of SDK client services such as
+    1. Service lookups
+    1. Session creation
+1. Horizontal scaling of circuit creation
+
+This means that for everything that SDK clients and tunnelers depend on, controllers
+can be scaled up and placed strategically to meet user demand.
```
qrkourier:
```suggestion
can be scaled out and placed strategically to meet user demand. 
```

docusaurus/docs/reference/ha/overview.md
qrkourier: Reading the limitations left me wondering what I should do with the information, probably because I don't interact directly with circuits in single-controller networks, and so I can't predict how these limitations would influence my deployment or configuration choices.

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+
+Flags:
+  -h, --help   help for cluster
+
+Use "ziti ops cluster [command] --help" for more information about a command.
+```
+
+## Growing the Cluster
+
+Once a single node is up and running, additional nodes can be added to it. They should be
```
dovholuknf: > Once a single node is up and running, additional nodes can be added to it

Maybe:
"After at least one controller is running as part of a cluster and initialized, additional nodes may be added to the cluster"

> They should be configured the same as the initial node, though they will have different addresses.

Additional nodes should be configured similarly as the initial node, referencing a different address (or something like that)?

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+
+Controller use the control channel listener to communicate with each other. Unlike
+routers, they need to know how to reach each other, so an advertise address must
+be configured.
+
+```yaml
+ctrl:
+  listener: tls:0.0.0.0:6262
+  options:
+    advertiseAddress: tls:192.168.1.100:6262
```
qrkourier: This doc uses IPv4 addresses from 192.168.1/24 and 192.168.3.0/24. It's unclear if it was intentional to imply the controllers are dual-homed. Still, I suggest using domain names to reduce cognitive load on the reader and emphasize their importance in operational scenarios.

Additionally, this is an opportunity to demonstrate and assume a single-port (ALPN) M.O.. 6262/TCP remains Ziti's internal default, at least for the in-built controller configuration generator, for the ctrl listener, but in practice, all Linux, Docker, and Kubernetes deployment packages and templates use 1280/TCP as the single, shared, default port for the controller's TLS servers.

```suggestion
    advertiseAddress: tls:ziti1.example.com:1280
```

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+
+Once a single node is up and running, additional nodes can be added to it. They should be 
+configured the same as the initial node, though they will have different addresses.
+
+The first node, as configured above, is running at `192.168.1.100:6262`. 
+
+If the second node is running at `192.168.1.101:6262`, then it can be added to the 
+cluster in one of two ways. 
+
+### From An Existing Node
```
dovholuknf: IMO the titles break up the flow here too much. Persaonlly, I'd rather see two bullets instead but -- whatever :) it's a nit...

Also "we" shows up here twice and should be reworked

docusaurus/docs/reference/ha/certificates.md
```diff
@@ -0,0 +1,86 @@
+---
+sidebar_label: Certificates
+sidebar_position: 20
+---
+
+# Controller Certificates
+
+For controllers to communicate and trust one another, they need certificates that have
+been generated with the correct attribute and relationships.
```
dovholuknf: This should expand on what SPIFFE IDs are and why they are used. I see them mentioned in the requirements section below, but adding the info to this first paragraph is useful. I believe many people will read the first paragraph and then skim the rest. Adding that sort of thing here helps them skim

docusaurus/docs/reference/ha/certificates.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Requirements
+
+1. The certificates must have a shared root of trust
+2. The controller client and server certificates must contain a 
+   [SPIFFE ID](https://spiffe.io/docs/latest/spiffe-about/spiffe-concepts/#spiffe-id)
+
+## Steps to Certificate Creation
+There are many ways to set up certificates, so this will just cover a recommended configuration.
```
dovholuknf: > recommended configuration 

is such a loaded term. I'd rather remove that and refer to this as an example configuration. Or -- with all the spec-reading I've been doing -- use the words "non-normative"... and feel fancy about it.... :)

docusaurus/docs/reference/ha/certificates.md
```diff
@@ -0,0 +1,86 @@
...
+
+The primary thing to ensure is that controllers have a shared root of trust. 
+A standard way of generating certs would be as follows:
+
+1. Create a self-signed root CA
```
dovholuknf: All root CA's are "self-signed" so it's technically redundant to use "self-signed root ca". The pedants out there might quibble...

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+```
+ 
+Finally, for sessions to work across controllers, JWTs are used. To enable these
+an OIDC endpoint should be configured.
+
+```yaml
+web:
+  - name: all-apis-localhost
+    bindPoints:
+      - interface: 127.0.0.1:1280
```
qrkourier:
```suggestion
      - interface: 0.0.0.0:1280
```

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+        address: 127.0.0.1:1280
```
qrkourier:
```suggestion
        address: ziti1.example.com:1280
```

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+ziti controller run ctrl1.yml
+```
+
+Once the controller is up and running, it will see that it is not yet initialized, and will pause
+startup, waiting for initialization. While waiting it will periodically emit a message:
+
+```
+[   3.323] WARNING ziti/controller/server.(*Controller).checkEdgeInitialized: the 
+Ziti Edge has not been initialized, no default admin exists.  Add this node to a 
+cluster using 'ziti agent cluster add tls:localhost:6262' against an existing
```
qrkourier:
```suggestion
cluster using 'ziti agent cluster add tls:127.0.0.1:1280' against an existing 
```

Prefer single-port MO and loopback IP to avoid ambiguity when localhost resolves only to ::1

docusaurus/docs/reference/ha/certificates.md
```diff
@@ -0,0 +1,86 @@
...
+1. Create an intermediate signing cert for each controller
+1. Create a server cert using the signing cert for each controller
+1. Create a client cert using the signing cert for each controller
+1. Make sure that the CA bundle for each server includes both the root CA and the intermediate CA
```
dovholuknf: > Make sure that the CA bundle for each server includes both the root CA and the intermediate CA for that server

This needs specifics. Open the controller config file, fine the 'ca' line, add the intermediary CA... Also this instruciton conflcts with our latest best practice of using 'only root cas' in the ca bundle. This might need discussion?

I would think the client should just send the leaf/intermediate chain when connecting though?

docusaurus/docs/reference/ha/bootstrapping.md
qrkourier, line 1: Should we insert the HA quickstart somewhere for educational purposes?

https://github.com/openziti/ziti/blob/main/doc/ha/quickstart.md

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+## Growing the Cluster
+
+Once a single node is up and running, additional nodes can be added to it. They should be 
+configured the same as the initial node, though they will have different addresses.
+
+The first node, as configured above, is running at `192.168.1.100:6262`. 
+
+If the second node is running at `192.168.1.101:6262`, then it can be added to the
```
qrkourier:
```suggestion
If the second node is running at `ziti2.example.com:1280`, then it can be added to the 
```

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+
+Use "ziti ops cluster [command] --help" for more information about a command.
+```
+
+## Growing the Cluster
+
+Once a single node is up and running, additional nodes can be added to it. They should be 
+configured the same as the initial node, though they will have different addresses.
+
+The first node, as configured above, is running at `192.168.1.100:6262`.
```
qrkourier:
```suggestion
The first node, as configured above, is running at `ziti1.example.com:1280`. 
```

docusaurus/docs/reference/ha/certificates.md
```diff
@@ -0,0 +1,86 @@
...
+   for that server
+
+Note that controller server certs must contain a SPIFFE id of the form
```
dovholuknf: this should be listed in the requirements under the spiffe id section

dovholuknf: also i think this is the first time i've seen trustdomain mentioned and that needs some clear explanation

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+If the second node is running at `192.168.1.101:6262`, then it can be added to the 
+cluster in one of two ways. 
+
+### From An Existing Node
+
+From a node already in the cluster, in this case our initial node, we can add the 
+new node as follows:
+
+```bash
+user@node1$ ziti agent cluster add tls:192.168.3.101
```
qrkourier:
```suggestion
user@node1$ ziti agent cluster add tls:ziti2.example.com
```

I may have overlooked the explanation, but at this point in the review, I'm wondering when the `tls:` prefix is required. It seems to be implied at some points in the Ziti system, such as when defining an interface bind address, and it's unclear if it would be implied here if omitted.

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+### From A New Node
+
+We can also ask the new node, which is not yet part of the cluster, to reach
+out to an existing cluster node and request to be joined.
+
+```
+user@node2$ ziti agent cluser add tls:192.168.3.100
```
qrkourier:
```suggestion
user@node2$ ziti agent cluser add tls:ziti1.example.com
```

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+## Shrinking the Cluster
+
+From any node in the cluster, nodes can be removed as follows:
+
+```
+user@node1$ ziti agent cluster remove tls:192.168.3.101
```
qrkourier:
```suggestion
user@node1$ ziti agent cluster remove tls:ziti2.example.com
```

docusaurus/docs/reference/ha/certificates.md
```diff
@@ -0,0 +1,86 @@
...
+So if your trust domain is `example.com` and your controller id is `ctrl1`, then your SPIFFE id
+would be:
+
+```
+spiffe://example.com/controller/ctrl1
+```
+
+**SPIFFE ID Notes:**
+
+* This ID must be set as the only URI in the `X509v3 Subject Alternative Name` field in the
```
dovholuknf: this is another requirement imo - much more than a 'note'

docusaurus/docs/reference/ha/certificates.md
```diff
@@ -0,0 +1,86 @@
...
+  certificate.
+* These IDs are used to allow the controllers to identify each during the mTLS negotiation.
+* The OpenZiti CLI supports creating SPIFFE IDs in your certs
```
dovholuknf: feels appropriate for the 'Example' section moreso than in the 'spiffe id notes'

docusaurus/docs/reference/ha/certificates.md
```diff
@@ -0,0 +1,86 @@
...
+    * Use the `--trust-domain` flag when creating CAs
+    * Use the `--spiffe-id` flag when creating server or client certificates
+
+## Example
+
+Using the Ziti PKI tool, certificates for a three node cluster could be created as follows:
```
dovholuknf: > Ziti PKI

The `ziti` CLI contains a subcommand for managing a PKI under `ziti pki`. This subcommand can be used to configure a three-node cluster using the following commands:

docusaurus/docs/reference/ha/certificates.md
qrkourier:
```suggestion
1. Create a root CA
```

It's fine, but redundant, since all roots are self-signed, and self-signed are roots by definition.

docusaurus/docs/reference/ha/certificates.md
dovholuknf: Also describing the topology would be nice. Maybe a table?

|Controller|IP|SPIFFE ID|
|--|--|--|
|ctrl1 | 192.168.3.100:port | controller/ctrl1|
|ctrl2 | 192.168.3.101:port | controller/ctrl2|
|ctrl3 | 192.168.3.102:port | controller/ctrl3|

Also imo, using 101, 102, 103 is easier than the 'off-by-one' ip... :) Maybe even give the full example with trust domains etc.

docusaurus/docs/reference/ha/migrating.md
```diff
@@ -0,0 +1,45 @@
+---
+sidebar_label: Migrating
+sidebar_position: 30
+---
+
+# Migrating Controllers
```
dovholuknf: I'd move this down in the list to the bottom. Sure there are some who will need this, but the doc shouldn't focus on them imo

docusaurus/docs/reference/ha/migrating.md
```diff
@@ -0,0 +1,45 @@
...
+
+A controller can be moved from standalone mode to HA mode. It can also be returned
+from HA mode back to standalone mode.
```
dovholuknf: > from HA mode back to standalone mode

Why bother allowing people to go back? Is there a reason to support it? Also any mutation doc should start with __BACK UP YOUR CONFIG/CONTROLLER__ :)

docusaurus/docs/reference/ha/routers.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Configuration
+
+Instead of specifying a single controller, you can specify multiple controllers
+in the router configuration.
+
+```yaml
+ctrl:
+  endpoints:
+    - tls:192.168.3.100:6262
```
qrkourier:
```suggestion
    - tls:ziti1.example.com:1280
```

docusaurus/docs/reference/ha/routers.md
```diff
@@ -0,0 +1,62 @@
...
+    - tls:192.168.3.101:6262
```
qrkourier:
```suggestion
    - tls:ziti2.example.com:1280
```

docusaurus/docs/reference/ha/routers.md
```diff
@@ -0,0 +1,62 @@
...
+    - tls:192.168.3.102:6262
```
qrkourier:
```suggestion
    - tls:ziti3.example.com:1280
```

docusaurus/docs/reference/ha/migrating.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Standalone to HA
+
+### Requirements
+First, ensure that the controller's certificates and configuration meet the requirements
+in [Bootstrapping](./bootstrapping.md).
```
dovholuknf: shouldn't this be pointing to certificates?

docusaurus/docs/reference/ha/routers.md
```diff
@@ -0,0 +1,62 @@
...
+
+By default these will be stored in a file named `endpoints` in the same directory
+as the router config file.
+
+However, the file can be customized using a config file settings.
+
+```yaml
+ctrl:
+  endpoints:
+    - tls:192.168.3.100:6262
```
qrkourier:
```suggestion
    - tls:ziti1.example.com:1280
```

docusaurus/docs/reference/ha/migrating.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Data Model Migration
+The controller's data can be imported in one of two ways:
```
dovholuknf: Is there a benefit from supporting two ways? It just seems like a maintenance issue to allow two different ways. Can we remove one or the other? Are there reasons why I would do one or the other? I'd rather give people one way to do it and document that one way. This makes me think there are reasons to support both -- and if there are still compelling reasons, the differences/positives/drawbacks need to be doc'ed

docusaurus/docs/reference/ha/migrating.md
dovholuknf: ooooh and now i caught the "and configuration" part.  Yeah this needs a lot more information overall... So much that I sort of don't even want to support migration.... I feel like this is a topic of discussion... :/

docusaurus/docs/reference/ha/routers.md
```diff
@@ -0,0 +1,62 @@
...
+are running in a cluster.
+
+The data model on the router is periodically snapshotted, so it doesn't need to
+be fully restored from a controller on every restart. 
+
+The location and frequency of snapshotting can be [configured](../configuration/router#edge).
+
+## Controller Selection
+
+When creating circuits, routers will chose the most responsive controller, based on latency.
```
qrkourier:
```suggestion
When creating [circuits](/learn/core-concepts/security/SessionsAndConnections.md#data-plane), routers will chose the most responsive controller, based on latency.
```

docusaurus/docs/reference/ha/operations.md
```diff
@@ -0,0 +1,96 @@
...
+terminate and expect to be restarted. This is so in memory caches won't be out of sync with
+the database which has changed.
+
+## Snapshot Application and Restarts
+
+If a controller is out of communcation for a while, it may receive a snapshot to apply, rather
+than a stream of events.
+
+If a controller receives a snapshot to apply after starting up, it will apply the snapshot and then
+terminate. This assumes that there is a restart script which will bring the controller back up after
```
qrkourier:
```suggestion
terminate. This assumes that there is a process manager to restart the controller after
```

docusaurus/docs/reference/ha/operations.md
```diff
@@ -0,0 +1,96 @@
...
+## Open Ports
+
+Controllers now establish connections with each other, for two purposes.
+
+1. Forwarding model updates to the leader, so they can be applied to the raft cluster
+2. raft communication
+
+Both kinds of traffic flow over the same connection.
+
+These connections do not require any extra open ports as we are using the control channel listener
```
qrkourier: This implies the ALPN for controller-to-controller is something new, not `ziti-ctrl`. What's the new one? This is important for operators who probe their deployments to verify the correct certs are presented on certain ports, for example.

docusaurus/docs/reference/ha/operations.md
```diff
@@ -0,0 +1,96 @@
...
+to listen to both router and controller connections. As part of the connection process the
+connection type is provided and the appropriate authentication and connection setup happens based on
+the connection type. If no connection type is provided, it's assumed to be a router.
+
+## System of Record
+
+In controller that's not configured for HA, the bolt database is the system of record. In an HA
```
qrkourier:
```suggestion
In a controller that's not configured for HA, the bolt database is the system of record. In an HA
```

docusaurus/docs/reference/ha/operations.md
```diff
@@ -0,0 +1,96 @@
...
+setup, the raft journal is the system of record. The raft journal is stored in two places, a
```
qrkourier:
```suggestion
setup, the Raft journal is the system of record. The Raft journal is stored in two places: a
```

docusaurus/docs/reference/ha/operations.md
```diff
@@ -0,0 +1,96 @@
...
+* raft.db - the bolt database containing raft journal entries
+* snapshots/ - a directory containing raft snapshots. Each snapshot is snapshot of the controller
+  bolt db
+* ctrl.db - the controller bolt db, with the current state of the model
+
+The location of all three is controlled by the raft/dataDir config property.
+
+```yaml
+raft:
+  dataDir: /var/ziti/data/
```
qrkourier: :pray: Thank you for using the FHS in examples.

docusaurus/docs/reference/ha/topology.md
```diff
@@ -0,0 +1,81 @@
+---
+sidebar_label: Topology
+sidebar_position: 60
+---
+
+# Controller Topology
+
+This document discuss considerations for how many controllers a network might
```
qrkourier:
```suggestion
This document discusses cluster size and member placement. 
```

docusaurus/docs/reference/ha/overview.md
```diff
@@ -0,0 +1,55 @@
...
+sidebar_position: 05
+---
+
+# Controller HA 
+
+## Overview
+
+OpenZiti controllers can be run in a cluster for high availablity and performance scaling.
+
+:::warning
```
dovholuknf: You can change this WARNING title fwiw
```
:::warning[Controller HA is still in beta]
```

image

dovholuknf: also -- I think Controller HA should be nested under Configuration where the other Controller config stuff is.

docusaurus/docs/reference/ha/topology.md
```diff
@@ -0,0 +1,81 @@
...
+A network manager will want to scale the number controllers based on client demand and may want to 
+place additional controllers geographically close to clusters of clients for better performance.
+
+## Voting vs Non-Voting Members
+
+Because every model update must be approved by a quorum of voting members, adding a large number of voting
+members can add a lot of latency to model changes. 
+
+If more controllers are desired to scale out to meet client needs, only as many controllers as are needed
+to meet availability requirements for mangement needs should be made into voting members.
```
qrkourier:
```suggestion
Scale out voting members to meet management needs and non-voting members to meet client needs.
```

docusaurus/docs/reference/ha/topology.md
```diff
@@ -0,0 +1,81 @@
...
+to take model updates, such as provisioning identities, adding/changes services and updating policies.
+
+Since a two node cluster will lose some functionality if either node becomes unavailable, a minimum
+of 3 nodes is recommended.
+
+### Clients
+
+The functionality that controllers provide to clients doesn't require any specific number of controllers.
+A network manager will want to scale the number controllers based on client demand and may want to 
+place additional controllers geographically close to clusters of clients for better performance.
```
qrkourier, line 25: A latency expressed as an order of magnitude in milliseconds (<1, 10, 100) would be helpful for understanding what's meant by "close." Etcd informs the reader that updates can only be accomplished as quickly as network and disk IO allow.

docusaurus/docs/reference/ha/topology.md
```diff
@@ -0,0 +1,81 @@
...
+This document discuss considerations for how many controllers a network might 
+need and how to place them geographically.
+
+## Number of Controllers
+
+### Management
+
+The first consideration is how many controllers the network should be able to lose without losing
+functionality. A cluster of size N needs (N/2) + 1 controllers active and connected to be able
+to take model updates, such as provisioning identities, adding/changes services and updating policies.
```
qrkourier, line 16: This appears to be alluding to a quorum of voting members, so it would be more clear to say that instead of raising the question "are active and connected controllers the same thing as voting members?"

docusaurus/docs/reference/ha/topology.md
```diff
@@ -0,0 +1,81 @@
...
+* 1 in ap-southeast-4 (non-voting)
+* 1 in ap-south-2 (non-voting)
+
+Assuming the leader is one of us-east-1 or us-west-2, model updates will only need to be accepted by 
+one relatively close node before being accepted. All other controllers will recieve the updates as well,
+but updates won't be gated on communications with all of them.
+
+**Alternate**
+
+For even faster updates at the cost of an extra controller, two controllers could be in us-east, one in us-east-1
```
qrkourier:
```suggestion
For even faster updates at the cost of an extra controller, two controllers could be in US Eastern DCs: one in us-east-1
```

docusaurus/docs/reference/ha/topology.md
```diff
@@ -0,0 +1,81 @@
...
+and one in us-east-2. The third member could be in the eu. Updates would now only need to be approved by two
```
qrkourier:
```suggestion
and one in us-east-2. The third member could be in the EU. Updates would now only need to be approved by two 
```

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+
+```yaml
+raft:
+  dataDir: /path/to/data/dir
+```
+The `dataDir` will be used to store the following:
+
+* `ctrl-ha.db` - the OpenZiti data model bbolt database
+* `raft.db` - the raft bbolt database
+* `snapshots/` - a directory to store raft snapshots
```
qrkourier:
```suggestion
* `snapshots/` - a directory to store Raft snapshots
```

docusaurus/docs/reference/ha/bootstrapping.md
```diff
@@ -0,0 +1,164 @@
...
+The controller requires a `raft` section.
+
+```yaml
+raft:
+  dataDir: /path/to/data/dir
+```
+The `dataDir` will be used to store the following:
+
+* `ctrl-ha.db` - the OpenZiti data model bbolt database
+* `raft.db` - the raft bbolt database
```
qrkourier:
```suggestion
* `raft.db` - the Raft bbolt database
```

docusaurus/docs/reference/30-configuration/controller.md
```diff
@@ -477,7 +477,7 @@ profile:
 
 The raft section enables running multiple controllers in a cluster.
 
-- `bootstrapMembers` - (optional) Only used when bootstrapping the cluster. List of initial clusters
+- `initialMembers` - (optional) Only used when bootstrapping the cluster. List of initial clusters
   members. Should only be set on one of the controllers in the cluster.
 - `commandHandler` - (optional)
     - `maxQueueSize` - (optional, 1000) max size of the queue for processing incoming raft log
```
qrkourier:
```suggestion
    - `maxQueueSize` - (optional, 1000) max size of the queue for processing incoming Raft log
```

docusaurus/docs/reference/ha/data-model.md
```diff
@@ -0,0 +1,135 @@
...
+      minutes)
+* Write characteristics
+    * Writes only happen from administrators
+    * Speed needs to be reasonable, but doesn't need to be blazing fast
+    * Write availability can be interrupted, since it primarily affects management operations
+    * Must be consistent. Write validation can’t happen with stale data. Don’t want to have to deal
+      with reconciling concurrent, contradictory write operations.
+* Generally involves controller to controller coordination
+
+Of the distribution mechanisms we looked at, RAFT had the best fit.
```
qrkourier:
```suggestion
Of the distribution mechanisms we looked at, Raft had the best fit.
```

docusaurus/docs/reference/ha/data-model.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Raft Resources
+
+For a more in-depth look at Raft, see
+
+* https://raft.github.io/
+* http://thesecretlivesofdata.com/raft/
+
+### RAFT Characteristics
```
qrkourier:
```suggestion
### Raft Characteristics
```

docusaurus/docs/reference/ha/data-model.md
```diff
@@ -0,0 +1,135 @@
...
+    * Good but not stellar performance
+* Reads
+    * Every node has full state
+    * Local state is always internally consistent, but maybe slightly behind the leader
+    * No coordination required for reads
+    * Fast reads
+    * Reads work even when other nodes are unavailable
+    * If latest data is desired, reads can be forwarded to the current leader
+
+So the OpenZiti controller uses RAFT to distribute the data model. Specifically it uses the
```
qrkourier:
```suggestion
So, the OpenZiti controller uses Raft to distribute the data model. Specifically, it uses the
```

docusaurus/docs/reference/ha/data-model.md
```diff
@@ -0,0 +1,135 @@
...
+[HashiCorp Raft Library](https://github.com/hashicorp/raft/).
+
+### Updates
+
+The basic flow for model updates is as follows:
+
+1. A client requests a model update via the REST API.
+2. The controller checks if it is the raft cluster leader. If it is not, it forwards the request to
```
qrkourier:
```suggestion
2. The controller checks if it is the Raft cluster leader. If it is not, it forwards the request to
```

docusaurus/docs/reference/ha/data-model.md
```diff
@@ -0,0 +1,135 @@
...
+   the leader.
+3. Once the request is on the leader, it applies the model update to the raft log. This involves
```
qrkourier:
```suggestion
3. Once the request is on the leader, it applies the model update to the Raft log. This involves
```

docusaurus/docs/reference/ha/data-model.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Runtime Data Characteristics
+
+Runtime data has different characteristics than the model data does.
+
+* Not necessarily shared across controllers
+* Reads **and** writes must be very fast
+* Generally involves sdk to controller or controller to router coordination
+
+Because writes must also be fast, RAFT is not a good candidate for storing this data. Good
```
qrkourier:
```suggestion
Because writes must also be fast, Raft is not a good candidate for storing this data. Good
```

docusaurus/docs/reference/ha/routers.md
```diff
@@ -0,0 +1,62 @@
...
+  endpoints:
+    - tls:192.168.3.100:6262
+    - tls:192.168.3.101:6262
+    - tls:192.168.3.102:6262
+```
+
+If the controller cluster changes, it will notify routers of the updated 
+controller endpoints. 
+
+By default these will be stored in a file named `endpoints` in the same directory
```
dovholuknf: Is this file expected to be entirely managed by the router or is there an expectation that the file can be modified by administrators? that shoudl be documented. whether the file is monitored or not should be mentioned. (i presume it requires a restart). i expect the file to be ancillary and just where the router stores some state between restarts. some details about the implementation are probably warranted.

docusaurus/docs/reference/ha/routers.md
```diff
@@ -0,0 +1,62 @@
...
+by default when running in standalone mode. 
+
+The router data model can be disabled on the controller using a config setting,
+but since it is required for HA, that flag will be ignored if the controllers
+are running in a cluster.
+
+The data model on the router is periodically snapshotted, so it doesn't need to
+be fully restored from a controller on every restart. 
+
+The location and frequency of snapshotting can be [configured](../configuration/router#edge).
```
dovholuknf: maybe: can be [configured using the `db` and `dbSaveIntervalSeconds`](../configuration/router#edge) ?

docusaurus/docs/reference/ha/routers.md
```diff
@@ -0,0 +1,62 @@
...
+  endpointsFile: /var/run/ziti/endpoints.yaml
+```
+
+In general, a router should only need one or two controllers to bootstrap itself,
+and thereafter should be able to keep the endpoints list up to date with help 
+from the controller.
+
+## Router Data Model
+
+In order to enable HA functionality, the router now receives a stripped down
```
dovholuknf: are these two paragraphs helpful? As an operator, I don't really know what any of this means tbh other than the last two paragraphs here

docusaurus/docs/reference/ha/routers.md
dovholuknf: are there any implications important to the user that the most responsive controller based on latency creates circuits? is that relevant to the endpoints at all? Are circuits not 'model'? Is it useful to refer to "router data model" at all from a doc pov? This section makes me ask myself - what does this mean to me when i'm operating a network

docusaurus/docs/reference/ha/operations.md
```diff
@@ -0,0 +1,96 @@
+---
+sidebar_label: Operations
+sidebar_position: 50
+---
+
+# Operating a Controller Cluster
+
+## Restoring from Backup
+
+To restore from a database snapshot, use the following CLI command:
```
dovholuknf: I don't first need to put the cluster into a maintenance mode? What if model updates happen? This operation takes precedence?

docusaurus/docs/reference/ha/operations.md
```diff
@@ -0,0 +1,96 @@
...
+successfully delivered the metrics message to a controller. The flag will then be set to true. So
+the first controller to get the metrics message is expected to deliver the metrics message to the
+events system for external integrators. The other controllers will have `doNotPropage` set to true,
+and will only use the metrics message internally, to update routing data.
+
+## Open Ports
+
+Controllers now establish connections with each other, for two purposes.
+
+1. Forwarding model updates to the leader, so they can be applied to the raft cluster
```
dovholuknf: is it just me or are these two things kind of the same to an operator?  They feel like "mostly" the same thing to me. I get the difference, but it all feels like "cluster-related messaging" to me. Separating them in a bulleted list doesn't seem necessary to me. Also, another "we". I would just state what field controls the port and that the controller will leverage ALPN, allowing the port to be set to the same as the REST apis.

docusaurus/docs/reference/ha/operations.md
```diff
@@ -0,0 +1,96 @@
...
+* ctrl.db
+
+An HA setup will have:
+
+* raft.db - the bolt database containing raft journal entries
+* snapshots/ - a directory containing raft snapshots. Each snapshot is snapshot of the controller
+  bolt db
+* ctrl.db - the controller bolt db, with the current state of the model
+
+The location of all three is controlled by the raft/dataDir config property.
```
dovholuknf: Feels like this should link to the relevant configuration section?

docusaurus/docs/reference/ha/operations.md
```diff
@@ -0,0 +1,96 @@
...
+
+```yaml
+raft:
+  dataDir: /var/ziti/data/
+```
+
+When an HA controller starts up, it will first apply the newest snapshot, then any newer journal
```
dovholuknf: >  When an HA controller starts up, it will first apply the newest snapshot,

Why does this process differ from the one outlined just above that states: 

> If a controller receives a snapshot to apply after starting up, it will apply the snapshot and then terminate.

I assume there's a reason but it makes me wonder what the differences are between these two processes

docusaurus/docs/reference/ha/topology.md
```diff
@@ -0,0 +1,81 @@
...
+The functionality that controllers provide to clients doesn't require any specific number of controllers.
+A network manager will want to scale the number controllers based on client demand and may want to 
+place additional controllers geographically close to clusters of clients for better performance.
+
+## Voting vs Non-Voting Members
+
+Because every model update must be approved by a quorum of voting members, adding a large number of voting
+members can add a lot of latency to model changes. 
+
+If more controllers are desired to scale out to meet client needs, only as many controllers as are needed
```
dovholuknf: Overall this sentence is hard to read and should be reworked. Maybe as Ken suggested, maybe somethign else you like but it's hard to read. :)

docusaurus/docs/reference/ha/topology.md
```diff
@@ -0,0 +1,81 @@
...
+reducing availability.
+
+### Example
+
+**Requirements**
+
+1. The network should be able to withstand the loss of 1 voting member
+1. Controllers should exist in the US, EU and Asia, with 2 in each region. 
+
+To be able to lose one voting member, we need 3 voting nodes, with 6 nodes total.
```
dovholuknf: more "we" to remove pls

docusaurus/docs/reference/ha/certificates.md
```diff
@@ -0,0 +1,86 @@
...
+1. Make sure that the CA bundle for each server includes both the root CA and the intermediate CA
+   for that server
+
+Note that controller server certs must contain a SPIFFE id of the form
+
+```
+spiffe://<trust domain>/controller/<controller id>
+```
+
+So if your trust domain is `example.com` and your controller id is `ctrl1`, then your SPIFFE id
```
dovholuknf: 'you/your' is like 'we' and can almost always be eliminated

docusaurus/docs/reference/ha/topology.md
```diff
@@ -0,0 +1,81 @@
...
+to meet availability requirements for mangement needs should be made into voting members.
+
+Additionally a having a quorum of controllers be geographically close will reduce latency without necessarily
+reducing availability.
+
+### Example
+
+**Requirements**
+
+1. The network should be able to withstand the loss of 1 voting member
```
dovholuknf: generally "one" and "two" vs "1" and "2". Also, I'd like to see you explain WHY two in each region? Why not three total voting members? Was there a rationale?

I'd also like to see guidance on when to use non-voting members. I naively thought the five would be three voters and two non-voters but that wasn't what was in the example

docusaurus/docs/reference/ha/certificates.md
qrkourier: A slightly more personable tone seems ideal for this sort of practical doc.

dovholuknf: This is reference doc. not a guide. You/your/we/personable tone is not appropriate in reference doc

docusaurus/docs/reference/ha/certificates.md
qrkourier: That's what I expected to see, that leaves must be presented with their issuer chains, not that intermediates must appear in the root CA bundle, since they're not roots, and treating them like roots defeats the purpose of an intermediate.

docusaurus/docs/reference/ha/overview.md
plorenz: Good to know about the warning title. 

There's a lot more here than configuration, so I disagree about moving it under configuration.

docusaurus/docs/reference/ha/overview.md
plorenz: fixed

docusaurus/docs/reference/ha/overview.md
plorenz: Added a link to the router next project

docusaurus/docs/reference/ha/overview.md
plorenz: add resource section for raft and glossary

docusaurus/docs/reference/ha/overview.md
plorenz: fixed

docusaurus/docs/reference/ha/overview.md
plorenz: Please add an issue and we can put it on the post-1.0 todo list.

docusaurus/docs/reference/ha/overview.md
plorenz: Clarified that a one node cluster is valid

docusaurus/docs/reference/ha/overview.md
plorenz: Moved discussion of non-voting members to a glossary

docusaurus/docs/reference/ha/overview.md
plorenz: Removed these links for now, since the pages are also gone

docusaurus/docs/reference/ha/bootstrapping.md
plorenz: Moving add/remove nodes to operations. Splitting the rest into;

* overview
* certs
* config
* initialization

docusaurus/docs/reference/ha/bootstrapping.md
plorenz: If you tell a joiner to join an existing cluster, the rest of the cluster needs to know how to dial the joiner. This information is provided by controllers whenever they dial other controllers. 
It's also possible for a joiner to be given an incorrect, but working address. For example, the advertise address may be a global address, but the joiner may get an IP address or an internal name. The controller should decide how it generally referenced.

docusaurus/docs/reference/ha/bootstrapping.md
plorenz: Tweaked this a bit, to make it less thought provoking, maybe better?

docusaurus/docs/reference/ha/bootstrapping.md
plorenz: Tweaked the working.

docusaurus/docs/reference/ha/bootstrapping.md
plorenz: Moving to operations

docusaurus/docs/reference/ha/operations.md
plorenz: Model updates that happen before the restore will be overwritten. Model updates that happen after the restore will be applied to the restored db, and may fail if they aren't valid.

docusaurus/docs/reference/ha/operations.md
plorenz: It's still ziti-ctrl. The controller uses a channel framework header to distinguish between router and ctrl connections.

docusaurus/docs/reference/ha/operations.md
plorenz: Changed 'after starting up' to 'after startup is complete', to make it clear that snapshots can be applied during startup without a restart, but once it's up and running a snapshot will result in restart.

docusaurus/docs/reference/ha/bootstrapping.md
plorenz: It should use the advertiseAddress there, so I switched it to `tls:ctrl1.ziti.example.com:6262`

docusaurus/docs/reference/ha/bootstrapping.md
plorenz: Added notes to clarify this

docusaurus/docs/reference/ha/bootstrapping.md
plorenz: It would not be filled in here

docusaurus/docs/reference/ha/certificates.md
plorenz: switched to example

docusaurus/docs/reference/ha/certificates.md
plorenz: Searched for and removed we(s)/you(r)(s)

docusaurus/docs/reference/ha/certificates.md
plorenz: Moved up to requirements

docusaurus/docs/reference/ha/migrating.md
plorenz: Added a big warning about backing up config/controller.

We allow people to go back because it doesn't require any extra code from us, and it's something someone might want to do.

docusaurus/docs/reference/ha/migrating.md
plorenz: Made a link for each to make it clearer.

docusaurus/docs/reference/ha/migrating.md
plorenz: It's very minimal code to support both, since they mostly do the same thing. 

I'll kick this over to ops again, because I don't know if it's worth supporting both.

docusaurus/docs/reference/ha/routers.md
plorenz: fixed

docusaurus/docs/reference/ha/routers.md
plorenz: fixed

docusaurus/docs/reference/ha/routers.md
plorenz: fixed

docusaurus/docs/reference/ha/routers.md
plorenz: fixed

docusaurus/docs/reference/ha/routers.md
plorenz: Added some more details.

docusaurus/docs/reference/ha/routers.md
plorenz: Toss up if this belongs here or in data model. It seemed router specific, so I put it here. 

I re-phrased it so it was circuits vs terminators rather than circuits vs 'data model'

docusaurus/docs/reference/ha/migrating.md
plorenz: Ops agreed we should remove it, so it's removed :+1:

docusaurus/docs/reference/ha/topology.md
plorenz: Added some examples using latency in millis

docusaurus/docs/reference/ha/topology.md
plorenz: tried to clean it up

docusaurus/docs/reference/ha/topology.md
plorenz: There are three voting members and three extra for geographical closeness to clients.

qrkourier (approved):

Issue #1006, opened 2025-01-13
[Proposal] Add intercept.v1 config type doc reference to /reference/config-types/ for consistency


nenkoru:
After a thorough documentation reading session, I think it might be worth to put an `intercept.v1` config-type reference to the `Config Types` section of the documentation.

This would improve the overall consistency because for now for me to get to the reference page for the `intercept.v1` I have to do `Learn -> Core Concepts -> Configuration Store -> Tunneler Config Type intercept.v1`
When with host.v1 I could just do
`Reference -> Config Types -> host.v1`

qrkourier:
relatedly, https://github.com/openziti/ziti-doc/issues/334#issuecomment-2015886560

Issue #334, opened 2023-01-10
rework 'core-concepts'


dovholuknf:
The core concepts should cover these itesms
* Overlay (controller/router/etc)
* Identity
* Config / Config Types
* Services
* Service Policies

qrkourier:
These items are missing and need to be added or this list should replace whatever's there now?

qrkourier:
Our config info is split between https://openziti.io/docs/reference/config-types/ and https://openziti.io/docs/learn/core-concepts/config-store/overview, so I've been flipping between those sections to jog my memory about the shape of intercept.v1 and host.v1 configs.

qrkourier:
Related to :point_up: PR fixes links to config type specs: https://github.com/openziti/ziti-doc/pull/871

qrkourier:
This is a todo for an all-at-once docs overhaul that will be defined at a future date. To clarify, the docs should cover at least, but not limited to, the list of topics in the description.

PR #1120, opened 2025-04-30
describe forwardAddressTranslations

scareything:
requires https://github.com/openziti/ziti/pull/3023

### DIFF
--- a/docusaurus/docs/reference/config-types/host.v1.md
+++ b/docusaurus/docs/reference/config-types/host.v1.md
@@ -43,6 +43,8 @@ properties to indicate what should be forwarded:
 * `allowedProtocols` - the list of allowed protocols. Valid values include `tcp` and `udp`
 * `forwardAddress` - flag indicating that the target address of the intercepted connection is to be
   used. Can only be set to true.
+* `forwardAddressTranslations` - the list of translations that specify which address to connect
+  to for a given target address.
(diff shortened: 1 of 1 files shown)


docusaurus/docs/reference/config-types/host.v1.md
```diff
@@ -43,6 +43,8 @@ properties to indicate what should be forwarded:
 * `allowedProtocols` - the list of allowed protocols. Valid values include `tcp` and `udp`
 * `forwardAddress` - flag indicating that the target address of the intercepted connection is to be
   used. Can only be set to true.
+* `forwardAddressTranslations` - the list of translations that specify which address to connect
+  to for a given target address.
```
dovholuknf: I feel like a link here and an anchor below to a header like "Foward Addresss Translation Example" would be useful. as is it feels a bit too 'flow-y'

qrkourier:
> Our config info is split between https://openziti.io/docs/reference/config-types/ and https://openziti.io/docs/learn/core-concepts/config-store/overview, so I've been flipping between those sections to jog my memory about the shape of intercept.v1 and host.v1 configs.

Would it be possible to address the config type documentation split at this time?

Re:
- https://github.com/openziti/ziti-doc/issues/1006
- https://github.com/openziti/ziti-doc/issues/334#issuecomment-2015886560

dovholuknf (approved):

PR #1158, opened 2025-06-16
fix broken social link

dovholuknf:
missed a slash


dovholuknf:
adding to https://github.com/openziti/ziti-doc/pull/1160

PR #1280, opened 2026-05-06
add scalar api docs

nicoalba:
add scalar api docs


docusaurus/openziti_versioned_docs/version-1.x/reference/developer/api/index.mdx
```diff
@@ -58,7 +58,7 @@ for clients to use the client API they must first [authenticate](../../../learn/
 obtain either a [partial or fully authenticated](../../../learn/core-concepts/security/sessions.md#full-vs-partial-authentication)
 [API Session](../../../learn/core-concepts/security/sessions.md).
 
-[Explore the latest edge client API Reference](./01-edge-client-reference.mdx)
+[Explore the latest edge client API Reference](/docs/openziti/reference/developer/api/edge-client-api-reference)
```
dovholuknf: we shouldn't use /docs like this unless there's some reason. was this 'necessary' for docusaurus reasons? why not use the @openziti (or whatever the alias is)

nicoalba: good point. fixed.

dovholuknf (approved):

PR #1294, opened 2026-05-19
scalar fixes

nicoalba:
scalar fixes


docusaurus/src/theme/Layout/index.tsx
```diff
@@ -1,15 +1,30 @@
...
-
+import { useLocation } from '@docusaurus/router';
 const starProps = {
     repoUrl: 'https://github.com/openziti/ziti',
     label: 'Star OpenZiti on GitHub',
 };
 
 export default function Layout({ children }: { children: ReactNode }): ReactNode {
+    const { pathname } = useLocation();
+    const isApiPage = /(api-reference|openapi-reference)$/.test(pathname);
```
dovholuknf, line 12: do we actually have inconsistent routes in use like this? (api-reference|openapi-reference)?

dovholuknf (approved):

docusaurus/src/theme/Layout/index.tsx
nicoalba: yes, openapi-reference is platform-doc's route and api-reference is frontdoor's route, and ziti-doc's routes end in api-reference

PR #1309, opened 2026-06-04
enhance versioned docs

nicoalba:


docusaurus/docusaurus.config.ts
```diff
@@ -295,6 +299,18 @@ const config: Config = {
                     repoUrl: 'https://github.com/openziti/ziti',
                     label: 'Star OpenZiti on GitHub',
                 },
+                versionBanners: [
+                    {
+                        pathPrefix: `${docsBase}/latest`,
+                        message: 'Latest (development) — features here may not be available in any released version. For stable documentation, see Active LTS (2.0.x).',
```
dovholuknf: Maybe: "This is the latest development documentation and may describe features not yet available in a long-term-stable (LTS) release. For new features and active support, see Active LTS (2.0.x)."

Also that needs to be a link to take the user to the doc

dovholuknf: I'd be better to refer to some variable too instead of the hardcoded "Active LTS (2.0.x)". Like by referencing the PluginConfig.versions if possible

dovholuknf: It would be cool to refer to the release strategy doc in here too if we can https://github.com/openziti/ziti/blob/main/RELEASE_POLICY.md

dovholuknf (approved):

